PR #6, opened 2024-10-17
Fix codespell hook

alexdewar:
It turns out that -L lets you provide a list of words to ignore on the command line. What we actually want is to provide a file with a list of words to be ignored. The correct option for this is -I.

See: https://github.com/ImperialCollegeLondon/python-template/pull/137


alexdewar:
Tagging @cc-a because I can't choose reviewers for this repo

cc-a:
Cheers @alexdewar

PR #26, opened 2024-11-11
Add Django GroupMember model to represent research groups directly.

### LINKED ISSUE #17: Add a GroupMember model
Add a new model that directly represents the concept of a research group.

Currently Coldfront has the concept of whether a user is a PI. This is mostly only used to determine if a user can create new projects. As an alternative to representing groups via project membership, this issue implements a direct representation of research groups via a dedicated model.

The implementation can be very simple at first and may be expanded later:
- `owner` field that is a `ForeignKey` to the user model.
- `member` field that is a `ForeignKey` to the user model.

HarmonicReflux:
# Description

- Introduces a new `GroupMember` model to provide a direct representation of research groups.
- Adds `owner` and `member` fields as ForeignKey references to the user model, designating a group owner and associated member.
- This initial implementation lays the groundwork for expanding research group functionality in future updates.

Notes to reviewers:
- Please verify that the `on_delete` behaviour is appropriate for the use case.
- Migrations are not yet included, as `django manage.py` is currently unavailable.

Fixes #17 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/models.py
+++ b/imperial_coldfront_plugin/models.py
@@ -1,1 +1,28 @@
 """Plugin Django models."""
+
+from django.conf import settings
+from django.contrib.auth.models import User
+from django.db import models
+
+
+class GroupMember(models.Model):
+    """Membership relationship within a group, associating an owner with a member.
+
+    This model stores relationships where each instance represents an ownership
+    connection between two users within a group, where `owner` is the user who
+    owns the membership (or perhaps manages the group) and `member` is the
+    associated user.
+
+    Attributes:
+        owner (ForeignKey): A reference to the user designated as owner, connected to
+            AUTH_USER_MODEL. Deletes related memberships when the owner is deleted.
+        member (ForeignKey): A reference to the user designated as member, connected to
+            AUTH_USER_MODEL. Deletes related memberships when the member is deleted.
+    """
+
+    owner: User = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    member: User = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        """Returns a string representation of the GroupMember instance."""
+        return f"Group owner is: {self.owner}\n - Group member is: {self.member}"
--- a/imperial_coldfront_plugin/models.py
+++ b/imperial_coldfront_plugin/models.py
@@ -1,7 +1,6 @@
 """Plugin Django models."""
 
 from django.conf import settings
-from django.contrib.auth.models import User
 from django.db import models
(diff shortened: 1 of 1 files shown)


HarmonicReflux:
Strangely, when I run `pytest`, I now encounter the following error. Previously, right after cloning the repository, `pytest` completed without issues.

Screenshot from 2024-11-11 17-06-45

imperial_coldfront_plugin/models.py
```diff
@@ -1 +1,28 @@
...
+
+    Attributes:
+        owner (ForeignKey): A reference to the user designated as owner, connected to
+            AUTH_USER_MODEL. Deletes related memberships when the owner is deleted.
+        member (ForeignKey): A reference to the user designated as member, connected to
+            AUTH_USER_MODEL. Deletes related memberships when the member is deleted.
+    """
+
+    owner: User = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    member: User = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
```
cc-a, line 24: I believe you can get around the mypy error by removing the `User` annotation on the variable. The annotation is incorect as the attribute is not an instance of `User` but an instance of `ForeignKey` which is mapped to `User`.

imperial_coldfront_plugin/models.py
```diff
@@ -1 +1,28 @@
...
+
+    def __str__(self) -> str:
+        """Returns a string representation of the GroupMember instance."""
+        return f"Group owner is: {self.owner}\n - Group member is: {self.member}"
```
cc-a, line 28: I'm actually not a big of fan of adding str methods to models so lets remove it.

cc-a (changes requested):
Thanks @HarmonicReflux. Small couple of tweaks.

cc-a (approved):
Thanks @HarmonicReflux. I'm going to go ahead and merge this and then deal with your questions.

imperial_coldfront_plugin/models.py
cc-a: So mypy embraces a sort of minimal approach to typing where it will automatically infer as much type information as possible - see https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html for details. You _can_ add explicit annotations as well but I'd usually hesitate to do that because:
- I think it detracts from readability rather than adding because it adds clutter and redundant information to the codebase. Mypy will tell you if it needs additional info to work out the type of something. It may sound odd but I think the exact type of something is usually a secondary consideration if using mypy to check that everything is consistent.
- Explicitly typing everywhere makes it difficult to update or refactor later.
- A good ide can provide you the type information on demand on those occasions it's required.

Issue #21, opened 2024-11-01
Add group management features to nav bar


cc-a:
#18 and #19 add some views to the function. These should be made accessible under the navbar preferably as a separate submenu.

The catch here is that Coldfront doesn't seem to provide an extension point for additional menu items. This issue therefore requires some consideration of how to extend Coldfront to facilitate this.

PR #30, opened 2024-11-14
Add group members view with permission checks

### LINKED ISSUE #18: Group membership view
Requires #17.

Add a view function to the UI that displays the members of a research group based on instances of the `GroupMember` model. A simple list is sufficient for the time being though we may add a table or additional functionality later.

View should accept a user_pk argument and the behaviour should be something like:
- Check if the logged in user is the same as referenced by `user_pk` or an admin. If not then return permission denied.
- Check if the user is a PI. If not then render a page that says they do not own a group.
- Look up `GroupMember` instances where `owner==user_pk`.
- Render name of group members.

HarmonicReflux:
# Description

- Implement `group_members_view` to display members of a research group based on the `GroupMember` model.
- Added permission checks to allow access only for the group owner or an admin.
- Included check for PI (Principal Investigator) status to ensure group ownership.
- Render `group_members.html` for authorised users; display `no_group.html` with a message for non-PI users.

Fixes #18, #34. 

## Type of change

## Key checklist

## Further checks


cc-a:
Based on our session yesterday you're hopefully now able to try this out in the browser @HarmonicReflux?

HarmonicReflux:
I am getting closer, but encounter this error int eh CI/CD tests:
failed_test

I import `admin` using the codebase we already have, i.e. like so
`from django.contrib import admin`
so not sure why the error is thrown.

cc-a:
I've taken the liberty of pushing a couple of commits. The first addresses the issue you reported by updating the settings used during tests to include the `django.contrib.auth` module. This came up now as we added some functionality using the admin module. An unfortunate clunky artifact on of how standalone Django apps work.

I then hit a subsequent issue with mypy that I was unable to resolve. I'm going to create a follow up issue to look into how to deal with these mypy issues properly but have disabled the mypy checks for now. Maybe I'm just missing something but I'm increasingly dubious of the merits of using mypy with Django.

HarmonicReflux:
Thanks for this. 

Well, as per out discussions in person and online, maybe its best to deactivate `mypy` for this project entirely .
I think insufficient support of non-primitive types in Python are a "limitation of the language on purpose". If they are more pain than good for this project, disabling them may be our best bet, but that is just my opinion. 

It is certainly worth looking into it in more detail so we have a good assessment of the technical debt we incur switching `mypy` off.

SaranjeetKaur:
A question: Is a new file generated in the `migrations/` folder each time `makemigrations.py` is run? For example, in this PR there is a file `migrations/0003_remove_groupmembership_owner_researchgroup_and_more.py`.

AdrianDAlessandro:
> A question: Is a new file generated in the `migrations/` folder each time `makemigrations.py` is run? For example, in this PR there is a new file `migrations/0003_remove_groupmembership_owner_researchgroup_and_more.py`.

Yes, if there are changes to the models then we need to make migrations, which are then saved in `migrations/`

cc-a:
@AdrianDAlessandro  Unfortunately you'll have to type the url in manually. Until we get #21 there is nowhere obvious to put the link.

cc-a (approved):
We'll need some tests for this but let's do that in follow up PR.

AdrianDAlessandro:
> @AdrianDAlessandro Unfortunately you'll have to type the url in manually. Until we get #21 there is nowhere obvious to put the link.

I went to `/group/1/members/` and it wasn't a valid URL. Am I doing something wrong there?

imperial_coldfront_plugin/views.py
```diff
@@ -1 +1,53 @@
 """Plugin views."""
+
+from django.contrib.auth import get_user_model
+from django.contrib.auth.decorators import login_required
+from django.http import HttpRequest, HttpResponse, HttpResponseForbidden
+from django.shortcuts import get_object_or_404, render
+
+from .models import GroupMembership
+
+User = get_user_model()
```
AdrianDAlessandro, line 10: This feels wrong... Shouldn't we be using `get_user_model()` directly each time?

imperial_coldfront_plugin/views.py
```diff
@@ -1 +1,53 @@
...
+                      If the user is not a PI, renders `no_group.html` with an
+                      appropriate message.
+
+    Raises:
+        Http404: If no user is found with the provided `user_pk`.
+    """
+    user = get_object_or_404(User, pk=user_pk)
+
+    if request.user != user and not request.user.is_superuser:
+        return HttpResponseForbidden("Permission denied")
```
AdrianDAlessandro, line 46: Why can only superusers can view this?

imperial_coldfront_plugin/views.py
```diff
@@ -1 +1,53 @@
...
+
+    if not user.userprofile.is_pi:
```
AdrianDAlessandro, line 48: Is this `userprofile` a field provided by coldfront?

cc-a:
@AdrianDAlessandro I think you need to go to `icl/group/1/members/` as all of the plugin urls are added to Coldfront under a prefix.

AdrianDAlessandro (changes requested):
Ok, I managed to work out I need to navigate to http://127.0.0.1:8000/icl/group/1/members/ and it worked for when I wasn't a PI. But after setting `is_pi` to True it failed with this error (I believe the same error mypy was showing):

```
...
  File "/usr/src/imperial_coldfront_plugin/imperial_coldfront_plugin/views.py", line 51, in group_members_view
    group_members = GroupMembership.objects.filter(owner=user)
...
FieldError at /icl/group/1/members/
Cannot resolve keyword 'owner' into field. Choices are: group, group_id, id, member, member_id
```

imperial_coldfront_plugin/views.py
HarmonicReflux: yes.

imperial_coldfront_plugin/views.py
HarmonicReflux: Only users and superusers have access to the view. 
A superuser (say, an administrator) should have access to the view independent whether they are a user, though.

imperial_coldfront_plugin/views.py
HarmonicReflux: By assigning `get_user_model()` to `User`, we establish a consistent reference to the user model throughout the module. This is useful if our module is extended or reused within the project, as we see in line 43. Assigning it to `User` at the module level ensures that the function is called only once and reduces redundant lookups.

Also, if multiple parts of the codebase refer to `User`, maintaining a single reference (`User`) is easier to manage than using `get_user_model()` everywhere.

Moreover, I think that in project-specific code (non-reusable apps), it is common and entirely safe to `assign get_user_model()` to `User` because the user model (`AUTH_USER_MODEL`) is stable and does not change dynamically during runtime.

HarmonicReflux:
In my opinion, this PR is good to go.

SaranjeetKaur (approved):
Thanks @HarmonicReflux!

AdrianDAlessandro:
> Ok, I managed to work out I need to navigate to http://127.0.0.1:8000/icl/group/1/members/ and it worked for when I wasn't a PI. But after setting `is_pi` to True it failed with this error (I believe the same error mypy was showing):
> 
> ```
> ...
>   File "/usr/src/imperial_coldfront_plugin/imperial_coldfront_plugin/views.py", line 51, in group_members_view
>     group_members = GroupMembership.objects.filter(owner=user)
> ...
> FieldError at /icl/group/1/members/
> Cannot resolve keyword 'owner' into field. Choices are: group, group_id, id, member, member_id
> ```

This error is still happening. I've opened a new issue to tackle it #47

PR #41, opened 2024-11-21
Add model for unix uid

### LINKED ISSUE #32: Add model for unix uid data
Requires #25 

Part of #31.

A key piece of identity data we need to include in the export to SoV is the unix [uid](https://en.wikipedia.org/wiki/User_identifier) value assigned to each user. In order to be able to export it we need to store it locally. We should add a model to do so.

Each uid value should be mapped to a single user via a one-to-one relationship. The uid value should be stored as an integer.

SaranjeetKaur:
# Description

This pull request introduces a new model called `UnixUID` to the `imperial_coldfront_plugin` application. The `UnixUID` model is designed to map a user to a unique identifier. The changes include the creation of a migration file and the addition of the new model to the existing models file.

Fixes #32

## Type of change

## Key checklist

## Further checks


SaranjeetKaur:
The steps that I have followed so far:
- Added the `UserID` model to `models.py`
- Then run `python makemigrations.py` (since, here it is equivalent to `python manage.py makemigrations`)
- Then run `pre-commit run --all-files` to make sure all checks pass
- All ok when running `pytest` too

Questions:
1. The [Django tutorial ](https://docs.djangoproject.com/en/5.1/intro/tutorial02/)mentions 3 steps to making model changes:
    - Change your models (in models.py).
    - Run [python manage.py makemigrations](https://docs.djangoproject.com/en/5.1/ref/django-admin/#django-admin-makemigrations) to create migrations for those changes.
    - Run [python manage.py migrate](https://docs.djangoproject.com/en/5.1/ref/django-admin/#django-admin-migrate) to apply those changes to the database.
    
    I am not sure how to run the third step in the absence of a `manage.py` file
    
2. How to deploy and view the changes to the app?

imperial_coldfront_plugin/models.py
```diff
@@ -29,3 +29,24 @@ class GroupMembership(models.Model):
         on_delete=models.CASCADE,
         related_name="group_memberships_set",
     )
+
+
+class UserID(models.Model):
```
cc-a, line 34: I would go for `UnixUID` here as it's less ambiguous.

imperial_coldfront_plugin/models.py
```diff
@@ -29,3 +29,24 @@ class GroupMembership(models.Model):
...
+            Deletes the related identifier when the user is deleted.
+        identifier (IntegerField): A unique identifier for the user.
+    """
+
+    user = models.OneToOneField(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="user_id",
+    )
+    identifier = models.IntegerField(default=0)
```
cc-a, line 52: Lets remove the default value here as we'll always need to provide it. Where uid's are concerned 0 is a special value that refers to the root user (or superadmin).

cc-a (changes requested):
Very nice. Just a couple of small tweaks.

imperial_coldfront_plugin/models.py
SaranjeetKaur: Done!

imperial_coldfront_plugin/models.py
SaranjeetKaur: I see. Done!

PR #46, opened 2024-11-28
Group invitations

### LINKED ISSUE #19: Membership invitation functionality
Requires #17.

Implement a workflow to add new members to a research group:
- PI wants to invite a new person to their group.
- They click on "add group member" or similar in the UI.
- They are presented with a form requesting an email address. They enter the email address of the person they want to add and press submit. They receive a confirmation that an invitation has been sent.
- An email is sent to the invitee. The email contains a single use link with an embedded token that when clicked creates a `GroupMember` instance for that user.

The main advantage of the email based workflow is that it will work even for users that have not accessed the system and so for whom there is no associated user account.

This should be implementable as 2 view functions/classes.
1) to handle the form and form submission. This should be relatively straightforward with the exception of generating the token for the email link. See below for further details.
2) to handle the invitation acceptance. This should validate the token (also see below), check that the user making the request matches the email in the token and then creates the `GroupMember` instance for that user..

A key aspect will be generating and validating the token. See below suggestion from ChatGPT on how to handle this. The token will need to encode at a minimum the user pk for the PI and the email of the user that is being invited.

<details>
<summary>Group Invitation Link </summary>
If you prefer not to create a dedicated model for storing tokens, you can use Django's built-in `SignedToken` and `TimestampSigner` utilities to generate and validate tokens. Here's an overview of how you could handle this:

1. **Generate a signed token**:
   ```python
   from django.core.signing import TimestampSigner

   def generate_token(user_email):
       signer = TimestampSigner()
       token = signer.sign(user_email)
       return token
   ```

2. **Send the invitation email**:
   ```python
   from django.core.mail import send_mail
   from django.urls import reverse
   from django.conf import settings

   def send_invitation(user_email):
       token = generate_token(user_email)
       invitation_link = settings.SITE_URL + reverse('validate_invitation', args=[token])
       send_mail(
           'Your Invitation',
           f'Please use the following link to sign up: {invitation_link}',
           settings.DEFAULT_FROM_EMAIL,
           [user_email],
       )
   ```

3. **Validate the token in a Django view**:
   ```python
   from django.core.signing import BadSignature, SignatureExpired, TimestampSigner
   from django.shortcuts import render, redirect
   from django.http import HttpResponse

   def validate_invitation(request, token):
       signer = TimestampSigner()
       try:
           user_email = signer.unsign(token, max_age=86400)  # Token valid for 1 day (86400 seconds)
           # Perform the desired action, e.g., create a user
           return redirect('success_page')
       except (BadSignature, SignatureExpired):
           return HttpResponse('Invalid or expired token', status=400)
   ```

4. **Add URL patterns for the validation view**:
   ```python
   from django.urls import path
   from .views import validate_invitation

   urlpatterns = [
       path('validate/<str:token>/', validate_invitation, name='validate_invitation'),
   ]
   ```

By using Django's `TimestampSigner`, you avoid creating a dedicated model for tokens while still ensuring that tokens are secure and have an expiration time.
</details>

jamesturner246:
# Description

Just want to check I'm on the right track before updating the DB and making tests.

* Log in and go to `http://localhost:8000/icl/invite_to_group/` to send an invite email (also displays on-page for now). 
* Click the link in the email, or on the page to get the token validated.
* Tokens timeout after a day.

You will need to configure a dummy smtp server to check the emails. I just put the following in `coldfront_development_environment/local_settings.py`, which dumps outbound emails to console:

```py
EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'                                                                                         
```

By the way, the group membership migrations are not made yet. Shall I do the honours?

Fixes #19 

## Type of change

## Key checklist

## Further checks


imperial_coldfront_plugin/views.py
```diff
@@ -1 +1,59 @@
...
+    token = signer.sign_object(
+        {
+            "inviter_pk": request.user.pk,
+            "invitee_email": invitee_email,
+        }
+    )
+
+    # Send invitation via email.
+    # TODO: get_host does not include protocol -- need an automagic way to get it.
+    invite_url = "http://" + request.get_host() + reverse("accept_invite", args=[token])
```
cc-a, line 28: You can use [HttpRequest.build_absolute_uri](https://docs.djangoproject.com/en/5.1/ref/request-response/#django.http.HttpRequest.build_absolute_uri).

cc-a (approved):
Looking good @jamesturner246. I suggest you rebase onto #30 which has an updated structure for the group membership models (plus the migrations).

cc-a:
I'm going to add the `console` email backend as the default in the development environment as there is point in having another value in development.

jamesturner246:
I'm wondering it it's worth submitting this as-is for now, and leave the database modification and subsequent testing to another PR. Not completely understanding of the `GroupMembership` either, same as @AdrianDAlessandro mentioned. pytest is behaving weird as  well, which might be a sign to move on until next Monday..

The code is changed from above to include an index page at `/`, which includes the gorup add form for now, as I saw no better place to put it yet. Some user validation is also performed: assuming you use Imperial's oauth2, or whatever is used, you should fill in your imperial email in the form (`james.turner@impe...` in my case), or be presented with forbidden on validation.

jamesturner246:
Group memberships are now saved to the DB. Should be ready to go @cc-a.

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/index.html
```diff
@@ -0,0 +1,8 @@
+<h1>Invite user to group</h1>
```
cc-a, line 1: Similarly should extend the base template.

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/invite_to_group.html
```diff
@@ -0,0 +1,11 @@
+<h1>Invitation sent to {{ invitee_email }}</h1>
```
cc-a, line 1: Similarly lets aim for something more user oriented and this should extend the base template.

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,93 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
     group_members = GroupMembership.objects.filter(owner=user)
 
     return render(request, "group_members.html", {"group_members": group_members})
+
+
+@require_POST
+@login_required
+def invite_to_group(request: HttpRequest) -> HttpResponse:
```
cc-a, line 68: We should use a Django form class as this will give us extra validation and error handling logic. The django docs on it are quite good - https://docs.djangoproject.com/en/5.1/topics/forms/#building-a-form.

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,93 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+
+
+@login_required
+def index(request: HttpRequest) -> HttpResponse:
+    """Render the index page.
+
+    Args:
+        request: The HTTP request object containing metadata about the request.
+    """
+    return render(request=request, template_name="imperial_coldfront_plugin/index.html")
```
cc-a, line 153: The usual Django pattern is to handle the display and processing of a form in a single view function. The example linked in the Django docs shows how it works.

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,93 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+
+    Args:
+        request: The HTTP request object containing metadata about the request.
+        token: The token that was sent to the invitee.
+    """
+    signer = TimestampSigner()
+
+    # Validate token.
+    try:
+        invite = signer.unsign_object(token, max_age=86400)
```
cc-a, line 120: Lets make the expiry time a setting.

cc-a (changes requested):
Looking good. A few refinements to get it over the line.

jamesturner246:
Had a go a making it  more user-friendly, and tided up a bit. You will need to define `TOKEN_TIMEOUT = 86400` in your local settings.

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,87 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
     group_members = GroupMembership.objects.filter(owner=user)
 
     return render(request, "group_members.html", {"group_members": group_members})
+
+
+@login_required
+def send_group_invite(request: HttpRequest) -> HttpResponse:
```
AdrianDAlessandro, line 67: This should be restricted to PIs, right?

I was able to generate an invite link when I wasn't the owner of a group which meant there was an error when going to that link because there was no group associated with that link. So that needs to be checked before the email is sent.

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,87 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+    """
+    signer = TimestampSigner()
+
+    # Validate token.
+    try:
+        invite = signer.unsign_object(token, max_age=settings.TOKEN_TIMEOUT)
+    except SignatureExpired:
+        return HttpResponseBadRequest("Expired token")
+    except BadSignature:
+        return HttpResponseBadRequest("Bad token")
```
AdrianDAlessandro, line 121: These could maybe also have a template associated with them. Although it's not an essential for this PR

imperial_coldfront_plugin/admin.py
```diff
@@ -2,7 +2,16 @@
...
 from django.contrib import admin
 
-from .models import GroupMembership
+from .models import GroupMembership, ResearchGroup
+
+
+class ResearchGroupAdmin(admin.ModelAdmin):
+    """Admin configuration for the ResearchGroup model."""
+
+    pass
```
AdrianDAlessandro, line 11: This should include a `list_display` to show the names of the groups in the admin page

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,87 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+            "The invite token is not associated with this email address."
+        )
+
+    # Update group membership in the database.
+    group = ResearchGroup.objects.get(owner__pk=invite["inviter_pk"])
+    GroupMembership.objects.create(group=group, member=request.user)
+
+    return render(
+        request=request,
+        context={"inviter": group.owner, "group": group},
```
AdrianDAlessandro, line 135: Currently shows as "ResearchGroup object (1)"

```suggestion
        context={"inviter": group.owner, "group": group.name},
```

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,87 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+
+    return render(
+        request,
+        "imperial_coldfront_plugin/send_group_invite.html",
+        {"form": form},
+    )
+
+
+@login_required
+def accept_group_invite(request: HttpRequest, token: str) -> HttpResponse:
```
AdrianDAlessandro, line 106: One detail from the original issue that seems to be missing is this: "it will work even for users that have not accessed the system and so for whom there is no associated user account".

So this view cannot be `login_required`. However, we still need to check against the user... Unless we're relying on the fact people will sign in with SSO after clicking the link and thus create an account?

AdrianDAlessandro (commented):
Overall works after I made sure I had created two users, one of which was a PI and the other that wasn't. I've made a couple of observations about some details, hopefully they make sense

imperial_coldfront_plugin/views.py
jamesturner246: It's a good point. The SSO email will need to be known upfront by the inviter if the invitee is required to eventually login with an Imperial account. @cc-a?

imperial_coldfront_plugin/views.py
jamesturner246: Changed s.t. `send_group_invite` is forbidden unless user is a group owner.

imperial_coldfront_plugin/admin.py
jamesturner246: Done.

jamesturner246:
This should be good to go, besides @AdrianDAlessandro's last comment about SSO email, which I'm not sure how to address.

imperial_coldfront_plugin/views.py
cc-a: @AdrianDAlessandro yes, we're relying on SSO here to create the account details and (with #51 merged) to get user uids.

cc-a (changes requested):
One last tweak @jamesturner246. Could you take out the index view. We'll handle creating links in #21.

imperial_coldfront_plugin/views.py
```diff
@@ -51,3 +61,90 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+
+    # Check the correct user is using the token.
+    if invite["invitee_email"] != request.user.email:
+        return HttpResponseForbidden(
+            "The invite token is not associated with this email address."
+        )
+
+    # Update group membership in the database.
+    group = ResearchGroup.objects.get(owner__pk=invite["inviter_pk"])
+    GroupMembership.objects.create(group=group, member=request.user)
```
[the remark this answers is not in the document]
jamesturner246: Done.

cc-a (approved):
Thanks @jamesturner246

AdrianDAlessandro:
Just a note, do we now need to add some settings to the other repo? Specifically:

```python
EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
TOKEN_TIMEOUT = 86400
``

@cc-a @jamesturner246

cc-a:
Good point @AdrianDAlessandro. I already added `EMAIL_BACKEND` as that is just a generic Django setting that we should have in development anyway. For the token settings I'm going to implement the pattern for importing plugin settings that was drafted in #42 but not merged.

PR #51, opened 2024-12-03
Get user uids from Microsoft Graph API

cc-a:
# Description

The original plan was to get uid values from LDAP but after discussion with the identity team they were able to add the uid data to the user data stored in Azure. That means we can now retrieve uid values from the Microsoft Entra Graph API user the access token from the login workflow. This has a number advantages over the previous LDAP implementation:

- simplicity (compare this PR with #42)
- no extra dependencies
- uses only 1 source for all identity data
- no additional settings

To test this:

- Clear the development database (`docker compose down -v`).
- Make sure the main branch of the development environment is up to date.
- Make sure you have the OIDC environment variables set.
- Start the server and login via the browser and Imperial SSO.
- Open a shell `docker compose exec app coldfront shell` and check a UnixUID instance was created for your user, e.g.:

```python
from imperial_coldfront_plugin.models import UnixUID
UnixUID.objects.get(user__username="yourusername")
```

Fixes #42 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/oidc.py
+++ b/imperial_coldfront_plugin/oidc.py
@@ -10,7 +10,7 @@ from .models import UnixUID
 
 
 def _update_user(user: User, claims: dict[str, Any]) -> None:
-    user.username = claims["preferred_username"].rstrip("@ic.ac.uk")
+    user.username = claims["preferred_username"].removesuffix("@ic.ac.uk")
     user.email = claims["email"]
     user.first_name = claims["given_name"]
     user.last_name = claims["family_name"]
--- a/imperial_coldfront_plugin/oidc.py
+++ b/imperial_coldfront_plugin/oidc.py
@@ -2,8 +2,11 @@
 
 from typing import Any
 
+import requests
 from django.contrib.auth.models import User
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+
+from .models import UnixUID
 
 
 def _update_user(user: User, claims: dict[str, Any]) -> None:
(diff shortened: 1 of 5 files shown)


imperial_coldfront_plugin/oidc.py
```diff
@@ -2,9 +2,12 @@
...
 
+import requests
 from django.contrib.auth.models import User
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 
+from .models import UnixUID
+
 
 def _update_user(user: User, claims: dict[str, Any]) -> None:
     user.username = claims["preferred_username"].rstrip("@ic.ac.uk")
```
AdrianDAlessandro, line 13: This should be
```suggestion
    user.username = claims["preferred_username"].removesuffix("@ic.ac.uk")
```

imperial_coldfront_plugin/oidc.py
```diff
@@ -53,4 +57,14 @@ def get_userinfo(
         """
         user_info = super().get_userinfo(access_token, id_token, payload)
         user_info["preferred_username"] = payload["preferred_username"]
+
+        # get user uid from Microsoft Graph API using the access token
+        # uid is stored under a custom attribute
+        response = requests.get(
+            "https://graph.microsoft.com/v1.0/me?$select=onPremisesExtensionAttributes",
```
AdrianDAlessandro, line 64: Should this URL be defined in the settings (eventually)?

AdrianDAlessandro (approved):
Great, this is indeed a lot simpler than the LDAP option! Sorry it took so long to review.

I am still getting that same bug were it's stripping off the last "a" from my username, so I've suggested the fix here.

imperial_coldfront_plugin/oidc.py
cc-a: It's not a setting per se in that it we wouldn't expect it to vary but it probably shouldn't be such a magic value.

PR #55, opened 2024-12-12
Add app settings file

cc-a:
# Description

Adds a file for configuring plugin specific settings. These are picked up by the plugin import mechanism of the development environment.

Note this PR should be paired with the `plugin-settings` branch of the development environment.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/settings.py
+++ b/imperial_coldfront_plugin/settings.py
@@ -1,0 +1,6 @@
+"""Settings for the Imperial Coldfront plugin.
+
+These are imported into the project level settings by the Coldfront plugin mechanism.
+"""
+
+INVITATION_TOKEN_TIMEOUT = 86400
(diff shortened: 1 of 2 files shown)


imperial_coldfront_plugin/settings.py
```diff
@@ -0,0 +1,6 @@
+"""Settings for the Imperial Coldfront plugin.
+
+These are imported into the project level settings by the Coldfront plugin mechanism.
+"""
+
+INVITATION_TOKEN_TIMEOUT = 86400
```
SaranjeetKaur, line 6: Does that mean the invite token will timeout after 24 hours?

jamesturner246 (approved):
LGTM :+1:

imperial_coldfront_plugin/settings.py
cc-a: It does. But actually it should probably be longer and we can make that clearer.

SaranjeetKaur: Ok, so what if the invite is sent at the end of a Friday, then the 24 hours are not very helpful - what is usually a minimal good number of hours for the timeout?

AdrianDAlessandro (commented):
Just found I had a pending comment on their PR from last month...

PR #95, opened 2025-01-14
Send emails in background

### LINKED ISSUE #89: Email sending infrastructure
Ideally email should be sent as a background task. This avoids issues with timeouts when trying to contact an email server and allows configuration of retry logic. Coldfront uses [Django Q](https://django-q.readthedocs.io/en/latest/)  for background tasks.

We should create a generic Django Q task that can be used to send arbitrary emails that can be triggered from view tasks/model signal handlers.

cc-a:
# Description

Implements a wrapper function to make it simple to send emails as a background task and refactors the one view that sends emails already.

I've added a configuration for Django Q to the development environment so make sure you've got the latest `main` if you want to run it locally.

Fixes #89 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -118,7 +118,7 @@ def send_group_invite(request: HttpRequest) -> HttpResponse:
                 [invitee_email],
                 "HPC Access Invitation",
                 "You've been invited to join the access group of "
-                f"{request.user.get_full_name()}\n\n"
+                f"{request.user.get_full_name()} ({request.user.email})\n\n"
                 f"Click the following link to accept the invitation:\n{invite_url}",
             )
 
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -3,7 +3,6 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
-from django.core.mail import send_mail
 from django.core.signing import BadSignature, SignatureExpired, TimestampSigner
 from django.http import (
     HttpRequest,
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -51,6 +51,7 @@ def pytest_configure():
             "django.contrib.auth.middleware.AuthenticationMiddleware",
         ],
         TOKEN_TIMEOUT=60,
+        Q_CLUSTER={"sync": True},
```
jamesturner246, line 54: What does this do?

imperial_coldfront_plugin/views.py
```diff
@@ -109,13 +109,12 @@ def send_group_invite(request: HttpRequest) -> HttpResponse:
...
-            # Send invitation via email.
-            send_mail(
-                "You've been invited to a group",
-                f"Click the following link to accept the invitation:\n{invite_url}",
-                request.user.email,
+            send_email_in_background(
                 [invitee_email],
+                "HPC Access Invitation",
+                "You've been invited to join the access group of "
+                f"{request.user.get_full_name()}\n\n"
```
jamesturner246, line 116: Maybe add the username or email in brackets. As one of many James Turners, I know it can be annoying confusing to be called on name alone.

jamesturner246 (approved):
Yep, fine. Just one minor thing.

tests/conftest.py
cc-a: It's a configuration setting for Django Q that tells it to run things synchronously. Needed for testing so that it doesn't try and talk to an actual job consumer.

imperial_coldfront_plugin/views.py
cc-a: As a Cave-Ayland this is not a problem I ever experience.

cc-a: Done

Issue #72, opened 2025-01-13
Allow group managers to send invites


cc-a:
Requires #70.

Adjust permission logic on `send_group_invite` to make view accessible by managers.

PR #98, opened 2025-01-14
Member terms conditions

cc-a:
# Description

Fixes #77

Fairly straightforward implementation. Turned the `accept_group_invite` view into a typical form handling view. Getting the view returns a page with the terms and conditions. Posting it with valid data creates the group membership and renders the confirmation page.

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking, back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (whatever its nature)

## Key checklist

- [X] All tests pass (eg. `python -m pytest`)
- [X] The documentation builds and looks OK (eg. `python -m sphinx -b html docs docs/build`)
- [X] Pre-commit hooks run successfully (eg. `pre-commit run --all-files`)

## Further checks

- [X] Code is commented, particularly in hard-to-understand areas
- [X] Tests added or an issue has been opened to tackle that in the future. (Indicate issue here: # (issue))

PR #99, opened 2025-01-15
Adding `is_manager` feild

### LINKED ISSUE #70: Add `is_manager` field to GroupMembership model.
Should be a simple required BooleanField. Default of false.

Sahil590:
# Description
Adding is_manager field

Fixes #70 and #72 (issue)
I have added the new field to the model

## Type of change

## Key checklist

## Further checks


cc-a (changes requested):
Implementation looks good but this seems to include the changes from #98.

Sahil590:
My bad, I was making this change while on that branch and switched to this one with the changes. It can merge after #98 is merge.

Sahil590:
Added the logic for the `send_group_invite` view since it was a small change

cc-a (approved):
LGTM.

Sahil590:
@cc-a the test is failing at this point, not sure how to fix this
image

cc-a (changes requested):
I'm going to be super picky and say that I'd prefer a linear history for the migrations. Rather than merging you can just delete the migrations added in this PR and `makemigrations` again and you should end up with a single migration.

imperial_coldfront_plugin/views.py
```diff
@@ -89,7 +89,12 @@ def check_access(request: HttpRequest):
 @login_required
 def send_group_invite(request: HttpRequest) -> HttpResponse:
     """Invite an individual to a group."""
-    if not ResearchGroup.objects.filter(owner=request.user).exists():
+    if (
+        not ResearchGroup.objects.filter(owner=request.user).exists()
+        and not GroupMembership.objects.filter(
+            member=request.user, is_manager=True
+        ).exists()
+    ):
```
jamesturner246, line 97: I think you're checking to see if a manager twice here: once with `.filter(owner=request.user).exists()`, and again with `is_manager=True`. Or am I misunderstanding?

jamesturner246 (changes requested):
See comment. I think you are checking group leadership twice.

Else looks good besides Chris' things.

imperial_coldfront_plugin/views.py
Sahil590: There are different checks, owner and is_manager field are not the same.

cc-a (approved):

jamesturner246 (approved):
LGTM :+1:

PR #101, opened 2025-01-16
Microsoft Graph API Client

### LINKED ISSUE #64: Entra API Infrastructure
There are various pieces of logic that will require us to search the Imperial College directory to discover and retrieve identity data. We already do this for accessing uid data in `oidc.py` however this needs to be generalised to contexts where there is not a logged in user.

cc-a:
# Description

This PR provides a API client class for retrieving identity data with the Microsoft Graph API.

The implementation uses [uplink](https://uplink.readthedocs.io/en/stable/#) to create the `MicrosoftGraphClient` class. This provides a simple interface to abstract over making direct HTTP requests. Where relevant the return data from the API is transformed into a more useful form before being returned by the client. Two client methods are implemented so far:

- user_profile - returns a subset of the identity data fields available for a user, the returned attributes may need to be expanded.
- user_uid - returns the unix uid value for a user.

The previous use of the graph api in `oidc.py` has been refactored to use the new client.

Not currently required but this implementation adds a new setting - `MICROSOFT_TENANT_ID` - that is not required for current functionality but will be in the future. The development environment `docker-compose.yml` has been updated to pass through the environment variable.

Fixes #64

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/microsoft_graph_client.py
+++ b/imperial_coldfront_plugin/microsoft_graph_client.py
@@ -6,6 +6,11 @@ from uplink import Consumer, get, response_handler
 
 
 def _get_app_access_token():
+    """Get an access token for the application to use the Microsoft Graph API.
+
+    Fetches an access token that is enabled for app-only access i.e. not on behalf of a
+    logged in user.
+    """
     tenant_id = settings.MICROSOFT_TENANT_ID
     response = requests.post(
         f"https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token",
--- a/imperial_coldfront_plugin/microsoft_graph_client.py
+++ b/imperial_coldfront_plugin/microsoft_graph_client.py
@@ -1,0 +1,61 @@
+"""Interface for interacting with the Microsoft Graph API."""
+
+import requests
+from django.conf import settings
+from uplink import Consumer, get, response_handler
+
+
+def _get_app_access_token():
+    tenant_id = settings.MICROSOFT_TENANT_ID
+    response = requests.post(
+        f"https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token",
+        data={
+            "grant_type": "client_credentials",
+            "client_id": settings.OIDC_RP_CLIENT_ID,
+            "client_secret": settings.OIDC_RP_CLIENT_SECRET,
+            "scope": "https://graph.microsoft.com/.default",
+        },
+    )
+    return response.json()["access_token"]
+
+
+def parse_profile_data(response):
+    """Parse the user profile data from the API response into a useful format."""
+    data = response.json()
+    extension_attributes = data.get("onPremisesExtensionAttributes", {})
+    return {
+        "job_title": data.get("jobTitle"),
+        "department": data.get("department"),
+        "company_name": data.get("companyName"),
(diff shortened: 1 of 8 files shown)


imperial_coldfront_plugin/microsoft_graph_client.py
```diff
@@ -0,0 +1,78 @@
...
+    provides easy to use data structures for the responses.
+    """
+
+    @response_handler(parse_profile_data)
+    @get(
+        "users/{username}@ic.ac.uk?$select=jobTitle,department,companyName,userType,onPremisesExtensionAttributes"
+    )
+    def user_profile(self, username: str):
+        """Get the profile data for a user."""
+        pass
```
jamesturner246, line 61: I can't help myself. This one gets a `pass` but the other doesn't...

imperial_coldfront_plugin/microsoft_graph_client.py
jamesturner246, line 1: Does it make sense to mypy type check in here?

imperial_coldfront_plugin/microsoft_graph_client.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @response_handler(get_uid_from_response)
+    @get("users/{username}@ic.ac.uk?$select=onPremisesExtensionAttributes")
+    def user_uid(self, username: str):
+        """Get the Unix uid for a user."""
```
jamesturner246, line 66: On that note, what normally goes in the function with uplink, if all is handled in the handler?

jamesturner246 (approved):
All good. Couple tiny things,but won't reject over them.

imperial_coldfront_plugin/microsoft_graph_client.py
cc-a: We're not currently mypy checking anywhere.

imperial_coldfront_plugin/microsoft_graph_client.py
cc-a: Nothing. The method is sort of just a place to hang decorators off of.

imperial_coldfront_plugin/microsoft_graph_client.py
cc-a: Pedant :stuck_out_tongue_closed_eyes:

cc-a: Done.

imperial_coldfront_plugin/oidc.py
```diff
@@ -64,15 +56,9 @@ def get_userinfo(
...
-        # uid is stored under a custom attribute
-        response = requests.get(
-            ENTRA_UID_ENDPOINT,
-            headers={"Authorization": f"Bearer {access_token}"},
-        ).json()
-        user_info["uid"] = int(
-            response["onPremisesExtensionAttributes"][ENTRA_UID_ATTRIBUTE]
-        )
+        graph_client = get_graph_api_client(access_token)
+        user_info["uid"] = graph_client.user_uid(username)
```
AdrianDAlessandro, line 63: Yea, this is cleaner.

AdrianDAlessandro (approved):
This seems to work well for me

PR #104, opened 2025-01-21
allowing managers to remove users

### LINKED ISSUE #92: Allow group managers to remove users
Requires #71.

Adjust permissions for the functionality introduced in #71 so that the relevant view function can be accessed by group managers.

Sahil590:
# Description
This allows managers to be able to view the members in the group and also be able to remove users in the group. There is also a check preventing them from deleting themselves from the group. 

Fixes #92 and #107 (issue)

## Type of change

## Key checklist

## Further checks


tests/test_views.py
```diff
@@ -286,6 +286,14 @@ def test_invalid_groupmembership(self, user_client):
...
         assert response.status_code == HTTPStatus.NOT_FOUND
 
+    def test_non_manager_cannot_access(self, no_manager_group, auth_client_factory):
+        """Test that a user who is not a manager cannot access the view."""
+        non_manager, group = no_manager_group
+        client = auth_client_factory(non_manager)
+        response = client.get(self._get_url())
+        assert response.status_code == HTTPStatus.FORBIDDEN
+        assert response.content == b"Permission denied"
+
```
cc-a, line 296: This is checking the same thing as `test_not_group_owner` (which could be renamed `test_not_group_owner_or_manager`). We need a test to show that managers _can_ access the view.

tests/conftest.py
```diff
@@ -166,3 +166,14 @@ def manager_in_group(user_factory, research_group_factory):
...
+
+@pytest.fixture
+def no_manager_group(user_factory, research_group_factory):
+    """Return a user who is not a manager in a research group."""
+    from imperial_coldfront_plugin.models import GroupMembership
+
+    manager = user_factory()
+    group, memberships = research_group_factory(number_of_members=0)
+    GroupMembership.objects.create(group=group, member=manager, is_manager=False)
+    return manager, group
```
cc-a, line 180: This fixture seems to be equivalent to just calling `research_group_factory(number_of_members=1)` so I don't think we need a dedicated fixture for this.

imperial_coldfront_plugin/views.py
```diff
@@ -189,7 +189,11 @@ def remove_group_member(request: HttpRequest, group_membership_pk: int) -> HttpR
     group_membership = get_object_or_404(GroupMembership, pk=group_membership_pk)
     group = group_membership.group
 
-    if request.user != group.owner and not request.user.is_superuser:
+    if (
+        request.user != group.owner
+        and not request.user.is_superuser
+        and not group_membership.is_manager
+    ):
```
cc-a, line 195: As implemented I think this would only allow a manager to remove themselves from a group (which is probably the one removal we should disallow).

Sahil590: I also just noticed if they are a manager, they can't access the `group_members_view` only a PI can should i create an issue for this?

cc-a: Ah, good catch. Yes please.

cc-a (changes requested):

tests/test_views.py
```diff
@@ -94,13 +94,6 @@ def test_get(self, pi_group, pi_client):
         assert response.status_code == HTTPStatus.OK
         assert isinstance(response.context["form"], GroupMembershipForm)
 
-    def test_manager_can_access(self, manager_in_group, auth_client_factory):
-        """Test that a group manager can access the view."""
-        manager, group = manager_in_group
-        client = auth_client_factory(manager)
-        response = client.get(self._get_url())
-        assert response.status_code == 200
-
```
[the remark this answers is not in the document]
Sahil590: I must have missed this

tests/test_views.py
```diff
@@ -262,8 +255,16 @@ def _get_url(self, group_membership_pk=1):
...
+        self, research_group_factory, auth_client_factory, user_client, pi_group
+    ):
+        """Test non group owner or manager cannot access the view."""
+        group, memberships = research_group_factory(number_of_members=1)
+        client = auth_client_factory(group.owner)
+        response = client.get(self._get_url())
+        assert response.status_code == HTTPStatus.FORBIDDEN
+        assert response.content == b"Permission denied"
+
         group_membership = pi_group.groupmembership_set.first()
```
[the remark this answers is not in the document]
Sahil590: I can't add that test until the manager is able to view the members because the `remove_group_member` redirects to the member's view page making the test fail.

imperial_coldfront_plugin/views.py
```diff
@@ -52,14 +52,18 @@ def group_members_view(request: HttpRequest, user_pk: int) -> HttpResponse:
...
+    ).exists()
+
+    if request.user != user and not request.user.is_superuser and not membership:
         return HttpResponseForbidden("Permission denied")
 
     if not user.userprofile.is_pi:
         return render(request, "no_group.html", {"message": "You do not own a group."})
 
     group_members = GroupMembership.objects.filter(group__owner=user)
-
```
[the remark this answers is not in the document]
Sahil590: Do you think the re-write of this should be similar to the remove_group_member logic, where they take the group_membership_pk instead of the user_pk

tests/test_views.py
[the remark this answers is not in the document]
cc-a: Put it in as an [xfail](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) and we'll take that out when addressing #107.

cc-a (approved):
Great. Looks good.

PR #105, opened 2025-01-21
Add search step for user invitation.

cc-a:
# Description

As part of the user invitation, it is now possible to now search by name or username for people to invite. Implemented via a new search view that then submits data to the invitation view.

There is still no link in the UI to begin the invitation process so you can get started by going to <http://localhost:8000/icl/user_search/>.

Not that the search results may contain some invalid options such as devices or rooms. #103 will correct this behaviour.

Fixes #78 

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [ X New feature (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking, back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (whatever its nature)

## Key checklist

- [X] All tests pass (eg. `python -m pytest`)
- [ ] The documentation builds and looks OK (eg. `python -m sphinx -b html docs docs/build`)
- [X] Pre-commit hooks run successfully (eg. `pre-commit run --all-files`)

## Further checks

- [X] Code is commented, particularly in hard-to-understand areas
- [X] Tests added or an issue has been opened to tackle that in the future. (Indicate issue here: # (issue))

Issue #78, opened 2025-01-13
Add users to HPC group via search


cc-a:
When adding a member of a ResearchGroup the UI should allow searching for members of Imperial. This will search for user data from the Microsoft graph API, present a list of options and send invites as required.

Issue #103, opened 2025-01-21
Filter eligible HPC users


cc-a:
When adding users we want to exclude those identities that should not be eligible for HPC accesse. At the moment search results from the Microsoft Graph API include entries referring to rooms, devices, guest users etc. For this reason we want to do some simple filtering on the search results to limit displayed results to those corresponding to end users that should be able to gain access.

PR #106, opened 2025-01-21
Add search step for user invitation.

cc-a:
# Description

Recreation of #105 

As part of the user invitation, it is now possible to now search by name or username for people to invite. Implemented via a new search view that then submits data to the invitation view.

There is still no link in the UI to begin the invitation process so you can get started by going to http://localhost:8000/icl/user_search/.

Not that the search results may contain some invalid options such as devices or rooms. https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/issues/103 will correct this behaviour.

Fixes https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/issues/78

Fixes #105 

## Type of change

## Key checklist

## Further checks


Sahil590 (approved):
Code looks good. I know the manager permissions is new, but you could add the permissions into this PR, or it can be tackled later. 

EDIT: Turns out I was using the wrong test user for this and it does work with managers and group owners

Sahil590:
I guess a nice to have is to restrict the search page to only superuser, PI, managers and group owners.

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/user_search.html
```diff
@@ -0,0 +1,33 @@
...
+  {% if search_results %}
+    <form action="{% url 'imperial_coldfront_plugin:send_group_invite' %}" method="post"">
+      {% csrf_token %}
+      <table>
+        <tr>
+          <th>Name</th>
+          <th>Email</th>
+          <th>Username</th>
+          <th>Department</th>
+          <th></th>
```
jamesturner246, line 16:
```suggestion
          <th>Select</th>
```

? Subjective, but it took me a quick second to see the radio button at first.

imperial_coldfront_plugin/microsoft_graph_client.py
```diff
@@ -65,6 +83,15 @@ def user_profile(self, username: str):
     def user_uid(self, username: str):
         """Get the Unix uid for a user."""
 
+    @response_handler(parse_profile_data_list)
+    @headers(dict(ConsistencyLevel="eventual"))
```
jamesturner246, line 87: :exploding_head:

jamesturner246 (approved):
Okay, user_search gets profile data from search query on the graph API, which is passed to send_group_invite as before. Cool. :+1:

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/user_search.html
cc-a: Very valid observation I'll leave it for the time being as the whole appearance needs a general pass.

Issue #93, opened 2025-01-13
View/UI for removal of group managers


cc-a:
Essentially the converse of #73. Should be able to extend/reuse much of the logic.

PR #108, opened 2025-01-22
Group managers UI

### LINKED ISSUE #73: View/UI for creation of group managers
Requires #70.

Add the elements required for a PI to make a member of their group a manager. This functionality should be visible/accessible only to a group's owner/admin.

Requires:
- A view function that takes a `GroupMembership` pk as an argument. Should check that the active user is the group owner/an admin and update the `is_manager` field of the `GroupMembership` instance.
- Add a hyperlink to the url for the above view function for each member's entry in `group_members_view`.

Sahil590:
# Description
Added two view functions for adding and removing users, respectively. 
I have added a link as the HTML page doesn't look the cleanest, so can change if needed. 

Fixes #73 #93 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/templates/group_members.html
+++ b/imperial_coldfront_plugin/templates/group_members.html
@@ -4,10 +4,13 @@
   <ul>
     {% for member in group_members %}
       <li>
-        {{ member.member.get_full_name }} - <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
-        {% if request.user.userprofile.is_pi %}
-          <a href="{% url 'imperial_coldfront_plugin:make_manager' member.pk %}">Make manager</a>
-          <a href="{% url 'imperial_coldfront_plugin:remove_manager' member.pk %}">Remove manager</a>
+        {{ member.member.get_full_name }} -
+        <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
+        {% if request.user.userprofile.is_pi or request.user == member.group_owner %}
+          | <a href="{% url 'imperial_coldfront_plugin:make_manager' member.pk %}">Make manager</a>
+        {% endif %}
+        {% if member.is_manager and request.user.userprofile.is_pi or member.group_owner %}
+          | <a href="{% url 'imperial_coldfront_plugin:remove_manager' member.pk %}">Remove manager</a>
         {% endif %}
       </li>
     {% endfor %}
--- a/imperial_coldfront_plugin/templates/group_members.html
+++ b/imperial_coldfront_plugin/templates/group_members.html
@@ -6,7 +6,7 @@
       <li>
         {{ member.member.get_full_name }} -
         <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
-        {% if is_manager == False %}
+        {% if is_manager == False and member.is_manager == False %}
           | <a href="{% url 'imperial_coldfront_plugin:make_manager' member.pk %}">Make manager</a>
         {% endif %}
         {% if member.is_manager and is_manager == False %}
--- a/imperial_coldfront_plugin/templates/group_members.html
+++ b/imperial_coldfront_plugin/templates/group_members.html
@@ -6,10 +6,10 @@
       <li>
         {{ member.member.get_full_name }} -
         <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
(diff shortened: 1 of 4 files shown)


imperial_coldfront_plugin/views.py
```diff
@@ -264,3 +264,61 @@ def get_active_users(request: HttpRequest) -> HttpResponse:
...
+    """
+    group_membership = get_object_or_404(GroupMembership, pk=group_membership_pk)
+    group = group_membership.group
+
+    if (
+        request.user != group.owner
+        and not request.user.is_superuser
+        and not GroupMembership.objects.filter(
+            group=group, member=request.user, is_manager=True
+        ).exists()
```
cc-a, line 285: It's a fair question whether managers should be able to make other managers but I'd be tempted to lean on the side of no.

imperial_coldfront_plugin/templates/group_members.html
```diff
@@ -5,6 +5,10 @@ <h1>Group Members</h1>
     {% for member in group_members %}
       <li>
         {{ member.member.get_full_name }} - <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
+        <a href="{% url 'imperial_coldfront_plugin:make_manager' member.pk %}">Make manager</a>
+        {% if member.is_manager %}
+          <a href="{% url 'imperial_coldfront_plugin:remove_manager' member.pk %}">Remove manager</a>
+        {% endif %}
```
cc-a, line 11: Given my below comment about preventing managers from adding/removing other managers, this only be visible to group owners.

imperial_coldfront_plugin/views.py
```diff
@@ -264,3 +264,61 @@ def get_active_users(request: HttpRequest) -> HttpResponse:
...
```
cc-a, line 315: Ditto.

cc-a (changes requested):
Thanks @Sahil590. Looking good. One tweak to the behaviour I think.

imperial_coldfront_plugin/templates/group_members.html
```diff
@@ -5,6 +5,10 @@ <h1>Group Members</h1>
     {% for member in group_members %}
       <li>
         {{ member.member.get_full_name }} - <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
+        {% if request.user.userprofile.is_pi %}
```
Sahil590, line 8: @cc-a I have added the check for pi here but not sure what the check for group owner would look like would it be 
`{% if request.user.userprofile.is_pi  or request.user.userprofile == group.owner%}`

cc-a: Ugh. Starting to bump up against the need to reimplement the view function again. I'd create a new variable `is_manager` that is determined in the view function and pass it into the template context to use to control the rendering.

imperial_coldfront_plugin/templates/group_members.html
```diff
@@ -4,7 +4,14 @@ <h1>Group Members</h1>
...
       <li>
-        {{ member.member.get_full_name }} - <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
+        {{ member.member.get_full_name }} -
+        <a href="{% url 'imperial_coldfront_plugin:remove_group_member' member.pk %}">Remove access</a>
+        {% if request.user.userprofile.is_pi or request.user == member.group_owner %}
+          | <a href="{% url 'imperial_coldfront_plugin:make_manager' member.pk %}">Make manager</a>
+        {% endif %}
+        {% if member.is_manager and request.user.userprofile.is_pi or member.group_owner %}
+          | <a href="{% url 'imperial_coldfront_plugin:remove_manager' member.pk %}">Remove manager</a>
+        {% endif %}
```
cc-a, line 14: I believe the only logic required here is to hide both of these if `is_manager` is true.

cc-a (changes requested):

imperial_coldfront_plugin/templates/group_members.html
Sahil590: Yeah, I was overthinking this one too much.

Sahil590: It also needs to hide the remove button if the member is not manager.

cc-a: I believe that only group owners, managers and superusers can access the view. Managers are the only ones that shouldn't see the links so it should be enough to check that.

cc-a (approved):

PR #110, opened 2025-01-23
Manager email notification

### LINKED ISSUE #74: Email notification on group manager creation/removal
Requires #89 and #70.

Essentially whenever the `is_manager` field of the `GroupMembership` model is changed there should be an associated email notification to group owner and member. This should be triggered by the view functions for creating/removing group managers.

Sahil590:
# Description

Used the `send_email_in_backround` function to send the manager changes to the group owner

Fixes #74  (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -281,7 +281,7 @@ def make_group_manager(request: HttpRequest, group_membership_pk: int) -> HttpRe
         request.user != group.owner
         and not request.user.is_superuser
         and not GroupMembership.objects.filter(
-            group=group, member=request.user, is_manager=True
+            group=group, member=request.user
         ).exists()
     ):
         return HttpResponseForbidden("Permission denied")
@@ -311,7 +311,7 @@ def remove_group_manager(
         request.user != group.owner
         and not request.user.is_superuser
         and not GroupMembership.objects.filter(
-            group=group, member=request.user, is_manager=True
+            group=group, member=request.user
         ).exists()
     ):
         return HttpResponseForbidden("Permission denied")
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -16,6 +16,7 @@ from django.urls import reverse
 from .forms import GroupMembershipForm, TermsAndConditionsForm, UserSearchForm
 from .microsoft_graph_client import get_graph_api_client
 from .models import GroupMembership, ResearchGroup
+from .policy import user_eligible_for_hpc_access
 from .utils import send_email_in_background
 
 User = get_user_model()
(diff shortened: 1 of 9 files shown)


imperial_coldfront_plugin/views.py
```diff
@@ -300,6 +300,12 @@ def make_group_manager(request: HttpRequest, group_membership_pk: int) -> HttpRe
     group_membership.is_manager = True
     group_membership.save()
 
+    send_email_in_background(
+        [group.owner.email],
+        "New group manager",
+        f"{group_membership.member.get_full_name()} has been made a manager of your group.",  # noqa: E501
+    )
+
```
cc-a, line 308: Both emails should go to both the group manager and the affected user.

cc-a (changes requested):
Let's also update the tests for the view functions to make sure the email is being sent. You can see an example of how to use the `mailoutbox` test fixture in `TestSendGroupInviteView.test_post_valid`.

imperial_coldfront_plugin/views.py
Sahil590: Forgot about the tests, doing them now

Sahil590:
Having some issues with the tests. It keeps detecting the `mailoutbox` has being empty, but it works fine in the other tests.

cc-a:
> Having some issues with the tests. It keeps detecting the `mailoutbox` has being empty, but it works fine in the other tests.

Hey @Sahil590. Had a look at this and I think it comes down to the `user_factory` fixture creating users without email addresses. Oddly the django `send_mail` command doesn't raise an error in if it's passed blank strings it just silently doesn't send an email. If you add the email data to the fixture that should get things working.

Sahil590:
> > Having some issues with the tests. It keeps detecting the `mailoutbox` has being empty, but it works fine in the other tests.
> 
> Hey @Sahil590. Had a look at this and I think it comes down to the `user_factory` fixture creating users without email addresses. Oddly the django `send_mail` command doesn't raise an error in if it's passed blank strings it just silently doesn't send an email. If you add the email data to the fixture that should get things working.

Thanks @cc-a should be fine now.

tests/conftest.py
```diff
@@ -80,13 +80,19 @@ def user_factory(django_user_model):
...
+        first_name=None,
+        last_name=None,
+        email=None,
     ):
         user = django_user_model.objects.create_user(
             username=username or random_string(),
             is_superuser=is_superuser,
             first_name=first_name or random_string(),
             last_name=last_name or random_string(),
+            email=f"{random_string()}@example.com",
```
cc-a, line 95: Would be good for this to follow the same pattern as the other arguments to `create_user`. Currently the `email` keyword argument from above is being ignored.

cc-a (approved):
Great. Could you just address the small comment I made then we can merge.

PR #111, opened 2025-01-23
Group expire

### LINKED ISSUE #80: Add `end_date` to GroupMembership model
Should be a straightforward addition as a required DateField - https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.DateField

### LINKED ISSUE #81: Prune expired GroupMemberships
Requires #80.

Would make sense to implement as a Django management command - https://docs.djangoproject.com/en/5.1/howto/custom-management-commands/.

Logic is fairly simple. Query for any GroupMemberships with an `end_date` in the past and delete them.

### LINKED ISSUE #91: Scheduled task for GroupMembership prune
Requires #81.

Coldfront uses [Django Q](https://django-q.readthedocs.io/en/latest/schedules.html) to handle scheduled tasks (see [Django Q Docs: Schedules](https://django-q.readthedocs.io/en/latest/schedules.html)).

The quirk here is that Django Q stores scheduled tasks within the app database. Pruning expired users is an essential function of the app so we want to make sure that the scheduled task for this is automatically created. The best way to do this is via a [data migration](https://docs.djangoproject.com/en/5.1/topics/migrations/#data-migrations) which will populate the required `Scheduled` instance into the database.

jamesturner246:
# Description

Automatically prune old expired group memberships.
Fixes #80 
Fixes #81
Fixes #91 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/models.py
+++ b/imperial_coldfront_plugin/models.py
@@ -48,14 +48,11 @@ class GroupMembership(models.Model):
         ResearchGroup,
         on_delete=models.CASCADE,
     )
-
     member = models.ForeignKey(
         settings.AUTH_USER_MODEL,
         on_delete=models.CASCADE,
     )
-
     is_manager = models.BooleanField(default=False)
-
     expiration = models.DateTimeField()
 
 
--- a/imperial_coldfront_plugin/models.py
+++ b/imperial_coldfront_plugin/models.py
@@ -41,6 +41,7 @@ class GroupMembership(models.Model):
         member (ForeignKey): A reference to the user designated as member, connected to
             AUTH_USER_MODEL. Deletes related memberships when the member is deleted.
         is_manager (BooleanField): A boolean value indicating if a member is a manager.
+        expiration (DateTimeField): The date and time when the membership expires.
     """
 
     group = models.ForeignKey(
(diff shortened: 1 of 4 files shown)


Sahil590 (approved):
Overall, the code looks fine and I have tested the expiry.

imperial_coldfront_plugin/models.py
```diff
@@ -41,19 +41,19 @@ class GroupMembership(models.Model):
...
         on_delete=models.CASCADE,
     )
-
     member = models.ForeignKey(
         settings.AUTH_USER_MODEL,
         on_delete=models.CASCADE,
     )
-
     is_manager = models.BooleanField(default=False)
+    expiration = models.DateTimeField(null=True, blank=True)
```
cc-a, line 56: Should be required, i.e. not nullable.

cc-a (changes requested):
Thanks @jamesturner246. Looks nice. Just need to make the field required which will probably have a few knock-on effects.

imperial_coldfront_plugin/models.py
jamesturner246: Done.

cc-a (approved):
Grand. Thanks @jamesturner246

PR #112, opened 2025-01-23
Automatic PI identification

### LINKED ISSUE #65: Implement PI Identification Logic
Requires #64

The existing self service portal contains logic to automatically detect whether a user is in a Senior Academic role and should have their own associated research group. The existing logic has been found to be robust and is implemented [here](https://github.com/ImperialCollegeLondon/rcs-api-website/blob/f67e8585c1ce302abae67e0b8efe7c3b3de2deb0/app/registration/routes.py#L134-L186).

The logic to transpose relies on some additional identity data that we don't currently :
- record_status - corresponds to "Extension Attribute 5"
- department
- status - corresponds to "Extension Attribute 6"

The checks from [these lines](https://github.com/ImperialCollegeLondon/rcs-api-website/blob/f67e8585c1ce302abae67e0b8efe7c3b3de2deb0/app/registration/routes.py#L164-L166) are not relevant to the new system.

cc-a:
# Description

A port of the PI identification logic from the old portal. At the moment this is a single function in the policy module (and tests) that will be used in later PRs.

Fixes #65 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/policy.py
+++ b/imperial_coldfront_plugin/policy.py
@@ -29,3 +29,44 @@ def user_eligible_for_hpc_access(user_profile):
             ),
         ]
     )
+
+
+PI_DISALLOWED_DEPARTMENTS = [
+    "External Users",
+    "Reach Out",
+    "Catering Services",
+    "Careers Service",
+    "Early Years Education Centre",
+    "CONTRACTOR",
+    "NONE",
+    "Campus Services",
+    "Guest Access",
+    "Union",
+    "Registry",
+    "Residential Services",
+    "Estates Division",
+]
+PI_ALLOWED_TITLES = ["Fellow", "Lecturer", "Chair", "Professor", "Reader", "Director"]
+PI_DISALLOWED_TITLE_QUALIFIERS = ["Visiting", "Emeritus", "Honorary"]
+
+
+def user_eligible_to_be_pi(user_profile):
+    """Assess eligibilty of a user be a Principal Investigator."""
+    job_title = user_profile["job_title"]
+    if any(
+        (
+            user_profile["record_status"] != "Live",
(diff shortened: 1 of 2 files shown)


imperial_coldfront_plugin/policy.py
```diff
@@ -29,3 +29,44 @@ def user_eligible_for_hpc_access(user_profile):
...
+    "NONE",
+    "Campus Services",
+    "Guest Access",
+    "Union",
+    "Registry",
+    "Residential Services",
+    "Estates Division",
+]
+PI_ALLOWED_TITLES = ["Fellow", "Lecturer", "Chair", "Professor", "Reader", "Director"]
+PI_DISALLOWED_TITLE_QUALIFIERS = ["Visiting", "Emeritus", "Honorary"]
```
jamesturner246, line 50: Visiting profs can't be PIs?

jamesturner246 (commented):
Are these categories, and the others in here, relatively constant, or does any of it rely on manual entry? Seems a little on the fragile side if so. A shame they can't just all be precomputed as extra graph API parameters, but 'spose that is what's been given..

imperial_coldfront_plugin/policy.py
cc-a: Only if they ask nicely.

cc-a:
> Are these categories, and the others in here, relatively constant, or does any of it rely on manual entry? Seems a little on the fragile side if so. A shame they can't just all be precomputed as extra graph API parameters, but 'spose that is what's been given..

All the data ultimately comes from Imperial's source of truth on identity data. We ultimately have to trust that that data is accurate. Reports from the platforms team are that this logic has been sufficiently robust for them so far.

Sahil590 (approved):
LGTM!

PR #113, opened 2025-01-24
Implement get_group_data view

### LINKED ISSUE #36: View for group data export to SoV
Requires #34.

Once we have models to store the required group we can create the view function that will be accessed by SoV.

The logic for the view function should be something like:
- Get all `ResearchGroup` instances.
- Get all members for each research group.
- Format the data in the template.

The format for the data the same as a [unix group file](https://www.cyberciti.biz/faq/understanding-etcgroup-file/). See the endpoint provided by the current self service platform for reference - https://api.rcs.imperial.ac.uk/v1.0/hpc/groups.

AdrianDAlessandro:
# Description

This adds a view to provide the /etc/groups endpoint for SoV

One thing I noticed when doing this is that there are no restriction on the ResearchGroup name. It needs to have the same restrictions as a unix group name (ie no spaces, possibly more). Should this be a new issue?

Fixes #36 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/urls.py
+++ b/imperial_coldfront_plugin/urls.py
@@ -29,6 +29,7 @@ urlpatterns = [
         name="remove_group_member",
     ),
     path("active_users/", views.get_active_users, name="get_active_users"),
+    path("groups/", views.get_group_data, name="get_group_data"),
     path("user_search/", views.user_search, name="user_search"),
     path(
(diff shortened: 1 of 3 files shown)


cc-a (approved):
Thanks. Looks good @AdrianDAlessandro

PR #116, opened 2025-01-30
Add group membership extension view function

### LINKED ISSUE #86: GroupMembership Extension view function
We need to provide the functionality to allow a group owner/manager wants to extend a group membership.

We need a view function that:
- takes a GroupMembership pk as an argument.
- checks that the accessing user is an owner/manager of the group in question (or a superadmin).
- on GET request - displays the user's information and the current expiry date as well as a form allowing them to specify the length of an extension.
- on POST request - updates the expiry date of the GroupMembership based on the form data and redirects to `group_members_view`.

This is a typical get/post pattern for a view with a form. The `user_search` view function is a simple example to pattern off of.

SaranjeetKaur:
# Description

This PR adds a new group membership extension view function using the group membership pk as an argument.

To do:
- [x] check that the accessing user is an owner/manager of the group in question (or a superadmin).
- [x] on GET request - display the user's information and the current expiry date as well as a form allowing them to specify the length of an extension.
- [x] on POST request - updates the expiry date of the GroupMembership based on the form data and redirects to `group_members_view`.
- [x] add unit test

Fixes #86 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/templates/imperial_coldfront_plugin/extend_membership.html
+++ b/imperial_coldfront_plugin/templates/imperial_coldfront_plugin/extend_membership.html
@@ -1,7 +1,10 @@
 {% extends "common/base.html" %}
 {% block content %}
-  <p>Your current membership is due to expire on {{ expiry_date }}.</p>
-  Would you like to extend it?
+  <h1>Extend Membership</h1>
+  <p>User/member information:</p>
+  <p>Member name: {{ group_membership.member }}</p>
+  <p>Group name: {{ group_membership.group }}</p>
+  <p>Current expiry date: {{ group_membership.expiration }}</p>
   <form method="post">
     {% csrf_token %} {{ form.as_p }}
     <button type="submit">Extend Membership</button>
--- a/imperial_coldfront_plugin/templates/imperial_coldfront_plugin/extend_membership.html
+++ b/imperial_coldfront_plugin/templates/imperial_coldfront_plugin/extend_membership.html
@@ -1,0 +1,9 @@
+{% extends "common/base.html" %}
+{% block content %}
+  <p>Your current membership is due to expire on {{ expiry_date }}.</p>
+  Would you like to extend it?
+  <form method="post">
+    {% csrf_token %} {{ form.as_p }}
+    <button type="submit">Extend Membership</button>
+  </form>
+{% endblock content %}
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -422,10 +422,5 @@ def group_membership_extend(
         form = GroupMembershipExtendForm()
 
     return render(
-        request,
-        "imperial_coldfront_plugin/extend_membership.html",
-        {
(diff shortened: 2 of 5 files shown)


imperial_coldfront_plugin/templates/imperial_coldfront_plugin/extend_membership.html
```diff
@@ -0,0 +1,9 @@
+{% extends "common/base.html" %}
+{% block content %}
+  <p>Your current membership is due to expire on {{ expiry_date }}.</p>
+  Would you like to extend it?
```
cc-a, line 4: Currently looks like this is addressed to the user that is being extended rather than the user doing the extension.

imperial_coldfront_plugin/views.py
```diff
@@ -366,3 +366,61 @@ def remove_group_manager(
...
+    # extension.
+
+    if request.method == "GET":
+        return render(
+            request,
+            "imperial_coldfront_plugin/extend_membership.html",
+            {
+                "group_membership": group_membership,
+                "group": group,
+            },
```
cc-a, line 407: Not currently passing the form in here.

imperial_coldfront_plugin/views.py
```diff
@@ -366,3 +366,61 @@ def remove_group_manager(
...
+        )
+
+    # on POST request - updates the expiry date of the GroupMembership based on the
+    # form data and redirects to group_members_view.
+
+    if request.method == "POST":
+        form = GroupMembershipForm(request.POST)
```
cc-a, line 414: You'll need a dedicated form for this I think.

cc-a (commented):
Definitely looking on the right track. I've put some comments, though you were probably aware of these already.

SaranjeetKaur:
might add some more unit tests for the view

tests/test_views.py
```diff
@@ -570,3 +572,53 @@ def test_successful_manager_removal(self, pi, pi_client, pi_group, mailoutbox):
...
+        response = client.get(self._get_url())
+        assert response.status_code == HTTPStatus.FORBIDDEN
+        assert response.content == b"Permission denied"
+
+    def test_group_owner(self, pi_client, pi_group):
+        """Test that the group owner can extend a group membership."""
+        group_membership = pi_group.groupmembership_set.first()
+        response = pi_client.get(self._get_url(group_membership.pk))
+        assert response.status_code == HTTPStatus.OK
+        assert isinstance(response.context["form"], GroupMembershipExtendForm)
```
cc-a, line 600: Need to also check that the group membership is included in the context as this is currently missing which causes info to be missing when the page is rendered.

cc-a (changes requested):
Thanks @SaranjeetKaur. Looking very nice, good catch on the manager extending their own membership corner case. In addition to my comment I think this should have some tests for the manager related behaviours i.e. managers can access and extend memberships except their own.

cc-a (approved):
Great. Thanks for this.

Issue #117, opened 2025-01-30
Create a template tag to check for group manager.


Sahil590:
The mange your group button in the navbar was broken from the change in #118. Since the link requires the group ID a custom template tag will need to be created to get this information to be HTML. [Docs ](https://docs.djangoproject.com/en/5.1/howto/custom-template-tags/) from Django about custom template tags.

PR #118, opened 2025-01-31
modifying the group memebers view function

### LINKED ISSUE #107: Manager role cannot access the members view
I also just noticed if they are a manager, they can't access the `group_members_view` only a PI can should i create an issue for this? Or is it the intentional

_Originally posted by @Sahil590 in https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/pull/104#discussion_r1923546604_
 

Currently, if they are in a manager they cannot access the members view

Sahil590:
# Description

This just changes the view function to take the group ID as an argument for the group members view function. 
As a result of this change some other view functions that redirect to the members view had to be changed so did the tests. =
One thing broken is the navbar link which i will fix later on in #117 . 

Fixes #107  (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/urls.py
+++ b/imperial_coldfront_plugin/urls.py
@@ -8,7 +8,7 @@ app_name = "imperial_coldfront_plugin"
 
 urlpatterns = [
     path(
-        "group/<int:user_pk>/members/",
+        "group/<int:group_gid>/members/",
         views.group_members_view,
         name="group_members",
     ),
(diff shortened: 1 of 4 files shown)


cc-a (changes requested):
Thanks for this @Sahil590. Looks much saner overall. I'm afraid I think we shoud use `group.pk` rather than `group.gid` for the view function argument though. Sorry that's going to be a pain to update.

PR #147, opened 2025-02-12
Remove usage of `UserProfile.is_pi`

### LINKED ISSUE #133: Update PI logic in check_access and get_active_users views
In some places we're still using `UserProfile.is_pi` to determine whether a PI has an access group but this is incorrect. Should check if they are owner of a `ResearchGroup` instance.

cc-a:
# Description

Previously we assumed that we could use the `UserProfile.is_pi` as a proxy for whether a user owned a ResearchGroup. This has no changed and this PR brings the existing view functions in line by checking directly whether a user owns a ResearchGroup instead.

Fixes #133 

## Type of change

- X ] Bug fix (non-breaking change which fixes an issue)

## Key checklist

## Further checks


SaranjeetKaur (approved):
Makes it easier to understand!

PR #150, opened 2025-02-14
Account for existing group members when adding new group members

### LINKED ISSUE #134: Existing group members
If a user is already a member of a group then:
- they should not appear in search results when adding users.
- it should not be possible to send them another invite.

cc-a:
# Description

If a user is already a member of a group then this PR means that they no longer appear in search results when adding new group members and prevents a new invitation being sent to them. This primarily a QoL change as the database schema already prevents them being added to another group.

Fixes #134 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/policy.py
+++ b/imperial_coldfront_plugin/policy.py
@@ -1,9 +1,10 @@
 """Policy functionality governing the eligibility of users access RCS systems."""
 
+from django.contrib.auth import get_user_model
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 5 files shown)


SaranjeetKaur (approved):
Nice one!

cc-a:
I've taken another pass at this as it occurred to me this should cover all the ways that users can have access i.e. if they are a member of a group, or a group owner or a superuser.

SaranjeetKaur (approved):

PR #156, opened 2025-02-21
:sparkles: Add GraphAPISearch backend.

### LINKED ISSUE #137: Implement user search as a search backend
Coldfront provides an extensible framework for supporting different search methods. We should implement the graph api based user search as a backend for use when adding users to a project.

https://github.com/ubccr/coldfront/blob/0b96bd391136d0a62ee1840e3cae111e6904b999/coldfront/core/user/utils.py#L12-L37

dalonsoa:
# Description

Adds a search backend for the GraphAPI using coldfront base `UserSearch`. Then this is used in the search view.

For now, this has been added directly to `views.py`, but it might be better to add it to a different place (eg. `microsoft_graph_client.py`). I've no added tests seems it seems a bit meaningless, having to mock everything within the overriding method for it to work.

Fixes #137 

## Type of change

## Key checklist

- [x] The documentation builds and looks OK (eg. `python -m sphinx -b html docs docs/build`)

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -63,7 +63,7 @@ class GraphAPISearch(UserSearch):
         found = graph_client.user_search(user_search_string)
         for user in found:
             user["source"] = self.search_source
-        return found
+        return list(filter(user_eligible_for_hpc_access, found))
 
 
 @login_required
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -33,7 +33,13 @@ from .forms import (
 )
 from .microsoft_graph_client import get_graph_api_client
 from .models import GroupMembership, ResearchGroup
-from .policy import user_eligible_for_hpc_access, user_eligible_to_be_pi
+from .policy import (
+    check_group_owner_manager_or_superuser,
+    check_group_owner_or_superuser,
+    user_already_has_hpc_access,
+    user_eligible_for_hpc_access,
+    user_eligible_to_be_pi,
+)
 
 User = get_user_model()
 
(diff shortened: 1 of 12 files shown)


imperial_coldfront_plugin/views.py
```diff
@@ -43,6 +44,23 @@
...
+    ) -> list[str]:
+        """Searchers for a single user.
+
+        Args:
+            user_search_string: The user string to look for.
+            search_by: (Unused) Fields to look into. This backend always looks into the
+                user's name or username.
+        """
+        graph_client = get_graph_api_client()
+        return graph_client.user_search(user_search_string)
```
cc-a, line 61: Data returned here isn't in the expected format for the search backend interface. You can see the expected format [here](https://github.com/ubccr/coldfront/blob/0b96bd391136d0a62ee1840e3cae111e6904b999/coldfront/core/user/utils.py#L61-L67). It's pretty similar but we should add the missing keys. You may need to pull extra data back from the graph api for the first/last names. I'm not sure how reliable splitting the full name would be.

cc-a (changes requested):
Looking like a good start. For now lets bring the hpc user eligibility filtering into the search backend as well. We may want to do something more sophisticated later but AFAIK for now we'll be applying the same filter everywhere.

Could you also please add the setting `ADDITIONAL_USER_SEARCH_CLASSES = ["imperial_coldfront_plugin.views.GraphAPISearch"]` as that will integrate the search backend with Coldfront and let us use it in other places.

cc-a (approved):
Great. Thanks @dalonsoa. I might open a follow up issue to look at supporting the `search_by` parameter properly but this is a point to merge this so far I think.

PR #176, opened 2025-03-06
Adds using batch endpoints in GPFS

### LINKED ISSUE #160: GPFS API Client - batch endpoints
All POST, PUT and DELETE endpoints of the GPFS API trigger asyncronous jobs to be perfomed by the cluster. These jobs may take some time to complete. The data returned by these endpoints includes an id for the created job. The status of the job can then be checked via a dedicated endpoint. See [API docs](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=api-asynchronous-jobs).

Our client needs to be capable of dealing elegantly with this scenario. A suggested specification:

- Implement an interface to the job status endpoint as a private method.
- Implement a generic wrapper/decorator that takes the job id returned by a batch endpoint and polls the job status endpoint (using exponential backoff). There should be a configurable timeout value after which, if confirmation of successful job completion has not been received, a custom exception is raised. In the case of the job failing to complete successfully a custom exception should be raised.
- Implement calls to batch endpoints using the wrapper/decorator.

We currently only need to implement an interface to one endpoint as a proof of concept (suggest the [fileset creation endpoint](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamefilesets-post)).

dalonsoa:
# Description

_Please include a summary of the change and which issue is fixed (if any). Please also
include relevant motivation and context. List any dependencies that are required for
this change._

Fixes #160 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -1,9 +1,41 @@
 """Interface for interacting with the GPFS API."""
 
+from functools import wraps
+
 import requests
 from django.conf import settings
-from uplink import Body, Consumer, get, post, returns
+from uplink import Body, Consumer, get, post, retry, returns
 from uplink.auth import BasicAuth
+from uplink.retry.stop import after_delay
+from uplink.retry.when import raises
+
+
+class ErrorWhenProcessingJob(Exception):
+    """Handles errors in asynchronous jobs."""
+
+
+class JobStillBeingProcessed(Exception):
+    """Indicates that a job is still being processed."""
+
+
+def process_jobid_status_request(f):
+    """Process the jobid status check, raising custom exceptions if needed."""
+
+    @wraps(f)
+    def wrapper(*args, **kwds):
+        response = f(*args, **kwds)
+        if not 200 <= response.status_code < 300:
+            raise response.raise_for_status()
+
+        job_data = response.json()["jobs"][0]
+        if job_data["status"] == "FAILED":
+            raise ErrorWhenProcessingJob()
+        elif job_data["status"] == "RUNNING":
+            raise JobStillBeingProcessed()
+
+        return response
+
+    return wrapper
 
 
 class GPFSClient(Consumer):
@@ -26,6 +58,8 @@ class GPFSClient(Consumer):
     def create_fileset(self, filesystemName: str, body: Body) -> dict:
         """Creates a new fileset in the requested filesystem."""
 
+    @retry(when=raises(JobStillBeingProcessed), stop=after_delay(60))
+    @process_jobid_status_request
     @get("jobs/{jobId}")
     def _get_job_status(self, jobId: int):
         """Query the status of a job."""
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -25,3 +25,7 @@ class GPFSClient(Consumer):
     @post("filesystems/{filesystemName}/filesets")
     def create_fileset(self, filesystemName: str, body: Body) -> dict:
         """Creates a new fileset in the requested filesystem."""
+
+    @get("jobs/{jobId}")
+    def _get_job_status(self, jobId: int):
+        """Query the status of a job."""
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -2,7 +2,7 @@
 
 import requests
 from django.conf import settings
-from uplink import Consumer, get, returns
+from uplink import Body, Consumer, get, post, returns
 from uplink.auth import BasicAuth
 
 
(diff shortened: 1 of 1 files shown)


dalonsoa:
@SaranjeetKaur , the only thing left is to figure out how to raise a custom exception when the job timesout.

@cc-a , it would be good if you could have a look to make sure we go in the right direction.

cc-a (commented):
Looks good to me. Didn't realise uplink had such sophisticated retry logic built in. @SaranjeetKaur happy to help with any questions if you pick this up.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -2,8 +2,57 @@
...
+        Returns:
+            True if the request should be retried, False otherwise.
+        """
+        if not 200 <= response.status_code < 300:
+            return False
+
+        job_data: dict = response.json()["jobs"][0]
+        if job_data["status"] == "FAILED":
+            raise ErrorWhenProcessingJob(job_data["result"])
+        elif job_data["status"] == "TIMEOUT":
```
dalonsoa, line 50: @SaranjeetKaur , I don't think this is a possible job status. The options indicated at the end of the [response section of the docs](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamefilesets-post#bl1adm_apiv2postfilesystemfilesets__title__9) are:
```
"status":"RUNNING | COMPLETED | FAILED"
Status of the job.
```
So, on this side, we need to:
- Check how long the job is taken to complete.
- If it takes more than `settings.GPFS_API_TIMEOUT`, raise the custom exception.

Possibly you need to create a custom stop condition by customising the [`after_delay` one](https://uplink.readthedocs.io/en/stable/dev/decorators.html#uplink.retry.stop.after_delay), and use the custom version, instead.

I guess we might want to cancel the job as well, using the [job delete endpoint](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-jobsjobid-delete), but that might be a refinement, I suppose.

dalonsoa:
I think this is now complete, if anyone wants to give it an informal look. I need to add some tests, but that would need to be next week, I'm afraid.

cc-a:
Just a quick couple of observations.
- I don't think we want to cancel jobs in the event of timeouts. The timeout is to trigger a human admin to follow up and make sure everything went ok.
- It would be nice to wrap the interface to any generate a background job response with the job status check. Then I can just call create fileset or whatever and get the benefit of the status checking.

dalonsoa:
> * It would be nice to wrap the interface to any generate a background job response with the job status check. Then I can just call create fileset or whatever and get the benefit of the status checking.

Should I use some kind of `async` function to do this? Or threads?

dalonsoa:
It seems that for using `async` we need a different client. It seems to be fully supported, but I'm unsure if there are any unwanted side effects. https://github.com/prkumar/uplink/blob/master/examples/async-requests/asyncio_example.py

cc-a:
Apologies @dalonsoa, I wasn't being clear. I think what I was thinking of was something like:

```python
class Client(Consumer):
    @check_job_status
    @other_uplink_decorators
    def endpoint_that_creates_a_gpfs_job(self, ...):
        ...

    def check_job_status(self, func):
        def wrapped(*args, **kwargs):
            job_id = func(*args, **kwargs)["job_id"]
            return self. job_status_endpoint(job_id)
        return wrapped

    @uplink_decorators
    def get_job_status(self, ...):
        ...
```

dalonsoa:
I'm not sure if that would work, but even if it does, calling `client.endpoint_that_creates_a_gpfs_job` is still going to be a blocking method and not running in the background. Until `self.job_status_endpoint` is done, the function won't return.

dalonsoa:
Ok, think this is ready, @cc-a . I won't be able to work on adding tests until Thursday morning, so if this is blocking your progress, @SaranjeetKaur @Sahil590 , I'd suggest we move on and merge, and open an issue for me about adding the relevant tests.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -1,9 +1,100 @@
...
+        """Checks if delay is beyond maximum timeout.
+
+        Raises:
+            JobTimeout: if delay is greater than the maximum delay.
+
+        Yields:
+            False, if delay has not reached the maximum.
+        """
+        while True:
+            delay = yield 0
```
cc-a, line 63: I always forget you can do this with `yield`.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -1,9 +1,100 @@
...
+            if self._max_delay < delay:
+                raise JobTimeout()
+            yield False
```
cc-a, line 66: Can we handle all of the exception raising in `check_job_status`? Not sure what happens with `after_delay` if it times out bet we could presumably detect and just raise the exception in `check_job_status`.

cc-a (approved):
Looks great. One small question but otherwise great stuff. Looks a bit tricky tests wise so maybe will go ahead and merge as I'm not sure they'll add a great deal of value anyway.

Issue #166, opened 2025-02-27
AD - Create a group


cc-a:
When a project with the correct resource assignment and the correct metadata is created this should create a group in AD.

PR #177, opened 2025-03-06
Add view function to create rdf allocations

### LINKED ISSUE #165: AD - interface infrastructure
LDAP infrastructure to talk to AD. Should be done as tasks via Django Q.

cc-a:
# Description

In addition to the basic ldap infrastructure for interacting with AD, this adds an administrator only view function that:
- provides a simple way of setting up the coldfront db objects required to represent a storage allocation on the RDF.
- creates a corresponding ldap security group in AD that will provide user access to the RDF allocation fileset.

Interacting with LDAP requires a role account with sufficient permissions but I don't have one I can share at the moment. You won't be able to set the environment variables required and directly test the LDAP functionality but you should still be able to load up and try out the view function. In the tests, LDAP_ENABLED is overridden to be true but the ldap client is mocked essentially just check that the ldap functions are called. Down the line we might be able to implement a local ldap server via docker compose for development/testing purposes.

Fixes #165 #166

## Type of change

## Key checklist

## Further checks


imperial_coldfront_plugin/settings.py
```diff
@@ -14,3 +14,16 @@
 GPFS_API_USERNAME = ENV.str("GPFS_API_USERNAME", default="")
 GPFS_API_PASSWORD = ENV.str("GPFS_API_PASSWORD", default="")
 GPFS_API_VERIFY = ENV.bool("GPFS_API_VERIFY", default=False)
+
+LDAP_USERNAME = ENV.str("LDAP_USERNAME", default="")
+LDAP_PASSWORD = ENV.str("LDAP_PASSWORD", default="")
+LDAP_URI = ENV.str("LDAP_URI", default="")
+LDAP_USER_OU = ENV.str(
```
dalonsoa, line 21: Probably, some of these environment variables with obscure names or purpose should have a minimal docstring explaining what they are about.

imperial_coldfront_plugin/views.py
```diff
@@ -493,3 +506,102 @@ def group_membership_extend(
...
+    ).values_list("value", flat=True)
+    id_numbers = [
+        int(PROJECT_ID_REGEX.match(group_id).groupdict()["number"])
+        for group_id in group_ids
+    ]
+    try:
+        new_project_number = max(id_numbers) + 1
+    except ValueError:
+        # no results
+        new_project_number = 1
```
dalonsoa, line 529: Maybe a question of personal taste, but I'm not a big fan of using exceptions for flow control in the case of simple scenarios like this one. I'd rather do:
```suggestion
    new_project_number = max(id_numbers) + 1 if id_numbers else 1
```

imperial_coldfront_plugin/views.py
```diff
@@ -493,3 +506,102 @@ def group_membership_extend(
...
+            )
+            AllocationAttribute.objects.create(
+                allocation_attribute_type=storage_quota_attribute_type,
+                allocation=rdf_allocation,
+                value=form.cleaned_data["size"],
+            )
+
+            rdf_id_attribute_type = AllocationAttributeType.objects.get(
+                name="RDF Project ID"
+            )
```
dalonsoa, line 582: This bit is duplicated; it's already above.
```suggestion
```

dalonsoa (commented):
I haven't try to run it, yet, but leave a couple of comments, for now.

imperial_coldfront_plugin/forms.py
```diff
@@ -42,3 +44,27 @@ class GroupMembershipExtendForm(forms.Form):
         label="Extend by (in days)",
         min_value=1,
     )
+
+
+def get_project_choices():
```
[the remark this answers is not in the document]
cc-a: Right. It's an admin only interface. By choosing the project they're choosing which pi to give the storage space to.

cc-a:
The markdown link check appears to be a transient issue will rerun in a bit to clear.

imperial_coldfront_plugin/views.py
```diff
@@ -493,3 +506,95 @@ def group_membership_extend(
...
+        if form.is_valid():
+            rdf_id_attribute_type = AllocationAttributeType.objects.get(
+                name="RDF Project ID"
+            )
+            group_id = get_next_rdf_project_id()
+            if AllocationAttribute.objects.filter(
+                allocation_attribute_type=rdf_id_attribute_type,
+                value=group_id,
+            ):
+                raise ValueError("RDF project with ID already exists.")
```
dalonsoa, line 552: Seeing the way `get_next_rdf_project_id` works, it is impossible for it to ever return an id that already exist, so this condition will never happen, as far as I can tell. I think this bit can be removed.

tests/test_views.py
```diff
@@ -788,3 +801,100 @@ def test_new_group_created(
...
+            data=dict(project=pi_project.pk, end_date=end_date, size=size),
+        )
+        assertRedirects(response, reverse("home"), fetch_redirect_response=False)
+
+        project_id = format_project_number_to_id(1)
+        from coldfront.core.allocation.models import (
+            Allocation,
+            AllocationAttribute,
+            AllocationUser,
+        )
```
dalonsoa, line 878: Unless there's a reason for it, move these import to the top of the function. I'm surprised that `ruff` didn't freakout with this...

dalonsoa (approved):
I've added a few comments, but no blockers. LGTM.

A couple of general comments, that I suppose will be addressed in the future:
- There's no link anywhere to access the allocation creation form; you need to use the direct URL for that.
- The view could do with some beautifying...
image

tests/test_views.py
cc-a: Huh. Yeah, why didn't ruff catch that...?

cc-a:
Thanks @dalonsoa. #183 exists for the beautification. I had to implement the navbar link within the development environment and forgot to put in a PR for that so I'll do that now.

PR #185, opened 2025-03-12
Improving RDF allocation UI

### LINKED ISSUE #183: Beautify `rdf_allocation_form.html`

Sahil590:
# Description
Using the bootstrap components to improve the look of the form and used the suggested date picker.

Fixes #183 

## Type of change

## Key checklist

## Further checks


cc-a (changes requested):
Looks like nice progress. I'm not convinced the form inputs are being properly styled however. They should look like in the bootstrap docs - https://getbootstrap.com/docs/4.0/components/forms/

dalonsoa (commented):
What happens if we just use `{{ form }}` or `{{ form|crispy }}`, if we use `crispy` forms? What is exactly to be gain by using a loop and unpaking the field information? 

Is a general question, not necessarily related to this PR - although it is relevant here.

cc-a:
@dalonsoa Ah, I for some reason didn't think that crispy forms was included as a Coldfront dependency but it is so yeah we should totally use it. Crispy is basically just an automatic way of doing all the Bootstrap related gubbins that is being done manually in this PR. Just `{{form}}` will render the form very simply but doesn't use any nice bootstrap stuff.

Sahil590:
Thanks, @dalonsoa, for pointing that out. Code looks cleaner.

dalonsoa (approved):
Nice!!

Sahil590:
Is there something else I need to add to the tests to load the template tag ?

dalonsoa:
I don't think so, but I'm not sure. Do tests pass locally?

Sahil590:
no they are not passing locally either

cc-a (approved):
Thanks @Sahil590. Looks good.

PR #188, opened 2025-03-18
GPFS client: set quotas

### LINKED ISSUE #182: Add interface to GPFS api to set quotas
We need to add an interface to set quotas for individual filesets. Confusingly this has to be done with the [filesystem quota endpoint](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamequotas-post). As with #180 we want a private method wrapped by a public method that provides a more useful interface. The public interface should take the following arguments in addition to the filesystem name:
- fileset_name - should be passed to `objectName` of the private method.
- block_quota - should be passed to both `blockSoftLimit` and `blockHardLimit` of the private method.
- files_quota - should be passed to both `filesSoftLimit` and `filesHardLimit` of the private method.

When the public method calls the private method it should additionally pass:
- `operationType="setQuota"`
- `quotaType="FILESET"`
- `filesGracePeriod="null"`
- `blockGracePeriod="null"`

SaranjeetKaur:
# Description

This pull request adding methods to set quotas on filesystems. The changes include the addition of a private method `_set_quota` and a public method `set_quota`.

Fixes #182 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -182,8 +182,10 @@ class GPFSClient(Consumer):
         Args:
             filesystem_name: Name of the filesystem where the quota will be set.
             fileset_name: Name of the fileset to set the quota.
-            block_quota: Block quota.
-            files_quota: Files quota.
+            block_quota: Number that specifies the block soft limit and hard
+                    limit. The number can be specified using the suffix K, M, G, or T.
+            files_quota: Number that specifies the inode soft limit and hard
+                    limit. The number can be specified using the suffix K, M, or G.
 
         Returns:
             The response after successfully setting the quota.
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -159,3 +159,44 @@ class GPFSClient(Consumer):
             path=path,
             permissions=permissions,
         )
+
+    @check_job_status
+    @json
+    @post("filesystems/{filesystemName}/quotas")
+    def _set_quota(
+        self,
+        filesystemName: str,
+        **data: Body,
+    ) -> requests.Response:
+        """Method (private) to set a quota in the requested filesystem."""
+
+    def set_quota(
+        self,
+        filesystem_name: str,
+        fileset_name: str,
+        block_quota: str,
+        files_quota: str,
+    ) -> requests.Response:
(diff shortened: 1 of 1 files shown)


SaranjeetKaur:
Some questions:

The description section in the [quotas docs](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamequotas-post) say `"The perfileset quota must be disabled to successfully complete this API command."` - what does it mean? Do we have to change anything somewhere?

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -159,3 +159,44 @@ def create_fileset(
...
+        block_quota: str,
+        files_quota: str,
+    ) -> requests.Response:
+        """Method (public) to set a quota in the requested filesystem.
+
+        Args:
+            filesystem_name: Name of the filesystem where the quota will be set.
+            fileset_name: Name of the fileset to set the quota.
+            block_quota: Block quota.
+            files_quota: Files quota.
```
cc-a, line 186: Could maybe add a bit more detail here about accepted formats for the data.

cc-a (approved):
Thanks @SaranjeetKaur. Looks great. One small comment that would be nice to have.

cc-a:
> The description section in the [quotas docs](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamequotas-post#bl1adm_apiv2postfilesystemquota__title__3) say `"The perfileset quota must be disabled to successfully complete this API command."` - what does it mean? Do we have to change anything somewhere?

Not sure. I assume this refers to some configuration on the GPFS side. It seems to work for us so I think it's fine.

> Following the [example](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamequotas-post#bl1adm_apiv2postfilesystemquota__title__10), I assigned `block_quota="1M"` and `files_quota="1K"`. Then in the gpfs gui for the `anothertest` fileset, I can see the `Hard limit` and `Soft limit` both set to `4.00 MiB` - how is this allocation being made?

I tried setting to 5MB and it ended up as 8MB so I think it only operates the quotas in increments of 4MB. Should be in the noise for expected data sizes so I don't think we need to address.

cc-a (approved):
Grand.

PR #190, opened 2025-03-19
GPFSClient: Add background task and extend view

### LINKED ISSUE #181: Add fileset creation to `add_rdf_storage_allocation` view.
Extend the view function so that in addition to exporting the AD group data, the GPFS fileset is also created and the quotas set appropriately. This should be executed via background task.

SaranjeetKaur:
# Description

Adds a background task function and updates the view to use this task.

Fixes #181 

## Type of change

## Key checklist

## Further checks


SaranjeetKaur:
@cc-a I am still trying to understand the view function. This is a initial trial - please feel free to glance through and share your thoughts.

imperial_coldfront_plugin/views.py
```diff
@@ -539,6 +540,28 @@ def add_rdf_storage_allocation(request):
...
 
     if request.method == "POST":
+        filesystem_name = request.POST.get("filesystem_name")
+        fileset_name = request.POST.get("fileset_name")
+        owner_id = request.POST.get("owner_id")
+        group_id = request.POST.get("group_id")
+        path = request.POST.get("path")
+        permissions = request.POST.get("permissions")
+        block_quota = request.POST.get("block_quota")
+        files_quota = request.POST.get("files_quota")
```
cc-a, line 550: So these values should come from a variety of places:
- filesystem_name - should be a setting
- fileset_name - will use the value of `group_id` below
- owner_id and group_id - will be "root" for now
- path - should be a prefix (given by a setting) with `group_id` appended
- permissions - should be a setting - "700" is a sensible guess
- files_quota - should be a setting - 1000 is ok for development
- block_quota - will use the value from the form accessible as `form.cleaned_data["size"]` after `form.is_valid()` below.

imperial_coldfront_plugin/views.py
```diff
@@ -539,6 +540,28 @@ def add_rdf_storage_allocation(request):
...
+            "imperial_coldfront_plugin.tasks.task_create_fileset_set_quota",
+            filesystem_name,
+            fileset_name,
+            owner_id,
+            group_id,
+            path,
+            permissions,
+            block_quota,
+            files_quota,
+        )
```
cc-a, line 563: Should be moved down to around the same place that as the LDAP task.

imperial_coldfront_plugin/views.py
```diff
@@ -592,6 +605,14 @@ def add_rdf_storage_allocation(request):
             )
             messages.success(request, "RDF allocation created successfully.")
 
+            if settings.GPFS_ENABLED:
+                gpfs_create_fileset_in_background(
+                    filesystem_name, fileset_name, owner_id, group_id, path, permissions
+                )
+                gpfs_set_quota_in_background(
+                    filesystem_name, fileset_name, block_quota, files_quota
+                )
```
cc-a, line 614: We'll need to create the fileset and set the quota as part of the same background task. They can't be separate tasks because we need to make sure the fileset creation completes successfully before setting the quota and getting that sort of control with background tasks is a bit elaborate.

imperial_coldfront_plugin/settings.py
```diff
@@ -24,6 +24,22 @@
...
+GPFS_PERMISSIONS = ENV.str("GPFS_API_PERMISSIONS", default="700")
+"""Permissions for the fileset."""
+
+GPFS_FILES_QUOTA = ENV.int("GPFS_API_FILES_QUOTA", default=1000)
+"""Quota for the fileset."""
+
+GPFS_ENABLED = bool(
+    GPFS_FILESYSTEM_NAME and GPFS_FILESET_PATH and GPFS_PERMISSIONS and GPFS_FILES_QUOTA
+)
+
```
cc-a, line 42: I'd base this on whether the url, password and username settings have been provided.

imperial_coldfront_plugin/views.py
```diff
@@ -539,8 +543,17 @@ def add_rdf_storage_allocation(request):
         return HttpResponseForbidden()
 
     if request.method == "POST":
+        filesystem_name = settings.GPFS_FILESYSTEM_NAME
+        owner_id = "root"
+        group_id = "root"
```
cc-a, line 548: Note that this gets overridden by `group_id` on L547.

imperial_coldfront_plugin/views.py
```diff
@@ -539,8 +543,17 @@ def add_rdf_storage_allocation(request):
...
+        fileset_name = group_id
+        path = f"{settings.GPFS_FILESET_PATH}{group_id}/"
+        permissions = settings.GPFS_PERMISSIONS
+        files_quota = settings.GPFS_FILES_QUOTA
+
```
cc-a, line 553: Probably don't need these values stored as variables up here as they will only be used once when calling the background task.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -202,3 +204,20 @@ def set_quota(
...
+
+def background_task(func):
+    """Wrapper to run a function in the background."""
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        """Run the function in the background."""
+        return async_task(func, *args, **kwargs)
+
+    return wrapper
```
cc-a, line 217: Rather than copying this from `ldap.py` maybe lets make it a utility function in `tasks.py`.

cc-a (changes requested):
Very close. Just a few tweaks.

cc-a (approved):

PR #192, opened 2025-03-25
Update to GPFS fileset creation process

cc-a:
# Description

After working the platforms team a bit there was some finessing to how we create filesets. There will be further items to implement.

This PR:

- adds a `faculty`, `department` and `group` level to the fileset directory structure.
- creates new fileset as dependent upon a presumed to exist faculty level fileset.
- implements a new APi endpoint for creating directories within a fileset and uses this to create the required directories for new dependent filesets.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -61,6 +61,8 @@ class RDFAllocationForm(forms.Form):
     """Form for creating a new RDF allocation."""
 
     project = forms.ChoiceField(choices=get_project_choices)
+    faculty = forms.CharField()
+    department = forms.CharField()
     end_date = forms.DateField(
         widget=forms.DateInput(attrs={"type": "date"}),
         validators=[MinValueValidator(timezone.now().date())],
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -224,7 +224,11 @@ class GPFSClient(Consumer):
         pass
 
     def create_fileset_directory(
-        self, filesystem_name: str, fileset_name: str, path: str
+        self,
+        filesystem_name: str,
+        fileset_name: str,
+        path: str,
+        allow_existing: bool = False,
     ):
         """Create a new directory within a fileset.
 
@@ -233,16 +237,23 @@ class GPFSClient(Consumer):
           fileset_name: name of the fileset in which to create the directory.
           path: location of the directory as a relative path w.r.t the fileset path,
             will create new directories recursively as required.
+          allow_existing: if True do not raise an error if the directory already exists.
         """
-        return self._create_fileset_directory(
-            filesystem_name,
-            fileset_name,
-            path,
-            user="root",
-            group="root",
-            permissions="750",
-            recursive=True,
-        )
+        try:
+            return self._create_fileset_directory(
+                filesystem_name,
+                fileset_name,
+                path,
+                user="root",
+                group="root",
+                permissions="750",
+                recursive=True,
+            )
+        except ErrorWhenProcessingJob as e:
+            task_data = e.args[0]
+            if allow_existing and task_data["exitCode"] == 6:
+                return task_data
+            raise
 
 
 class FilesetCreationError(Exception):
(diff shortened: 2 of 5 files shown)


tests/test_views.py
```diff
@@ -880,13 +889,24 @@ def test_post(
...
         )
+
+        path = Path(
+            settings.GPFS_FILESET_PATH,
+            settings.GPFS_FILESYSTEM_NAME,
+            faculty,
+            department,
+            pi_project.pi.username,
+            project_id,
+        )
```
SaranjeetKaur, line 900: Interesting to see how this is expanded to include more info!

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -205,6 +211,39 @@ def set_quota(
...
+          path: location of the directory as a relative path w.r.t the fileset path,
+            will create new directories recursively as required.
+        """
+        return self._create_fileset_directory(
+            filesystem_name,
+            fileset_name,
+            path,
+            user="root",
+            group="root",
+            permissions="750",
```
SaranjeetKaur, line 243: How does "750" differ from "700" in permissions?

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -205,6 +211,39 @@ def set_quota(
...
+            recursive=True,
```
SaranjeetKaur, line 244: What happens if we set it to `False`?

SaranjeetKaur (approved):
Thanks @cc-a! Looks good. I haven't tried to recreate this locally, but will like to when I am back to office.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -205,6 +211,39 @@ def set_quota(
...
+          fileset_name: name of the fileset in which to create the directory.
+          path: location of the directory as a relative path w.r.t the fileset path,
+            will create new directories recursively as required.
+        """
+        return self._create_fileset_directory(
+            filesystem_name,
+            fileset_name,
+            path,
+            user="root",
+            group="root",
```
dalonsoa, line 242: Are these values placeholders or will always be the same?

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -219,16 +258,33 @@ def _create_fileset_set_quota(
...
     client = GPFSClient()
+    try:
+        client.create_fileset_directory(
+            filesystem_name,
+            parent_fileset,
+            str(path.parent.relative_to(path.parents[2])),
+        )
+    except ErrorWhenProcessingJob:
+        # will fail if directory already exists
+        pass
```
dalonsoa, line 277: Do we want to silently ignore these errors? There might be other reasons why this fails, I guess... At the very least, I would log the error somewhere.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -219,16 +258,33 @@ def _create_fileset_set_quota(
     fileset_name: str,
     owner_id: str,
     group_id: str,
-    path: str,
+    path: Path,
```
dalonsoa, line 261: It might be more elegant, and require less assumptions on the specific structure of the `path`, which might change in the future, to provide 2 arguments for it instead of only one:
- `department_path`: up to, and including, the department.
- `project_path`: the project-specific part of the path, including the PI. 

This way you can use one or the other, or a combination of both if needed `path = department_path / project_path`, as needed, and you are clearer of the overall structure of the paths.

dalonsoa (approved):
I've a couple of questions and a suggestion for a different way of dealing with the paths, but it looks good, otherwise.

imperial_coldfront_plugin/forms.py
```diff
@@ -61,6 +61,8 @@ class RDFAllocationForm(forms.Form):
     """Form for creating a new RDF allocation."""
 
     project = forms.ChoiceField(choices=get_project_choices)
+    faculty = forms.CharField()
+    department = forms.CharField()
```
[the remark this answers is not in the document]
cc-a: Agreed.

imperial_coldfront_plugin/gpfs_client.py
cc-a: Unsure at this point.

imperial_coldfront_plugin/gpfs_client.py
cc-a: Good shout. I've added a check on the return value to check for the reason. Have also moved the logic up into the client.

imperial_coldfront_plugin/gpfs_client.py
cc-a: Likewise a good shout.

imperial_coldfront_plugin/gpfs_client.py
cc-a: So in the case of `/dir_a/dir_b` recursive tells it to additionally create `dir_a` if it doesn't exist otherwise we'd get an error that it can't create `dir_b` because `dir_a` doesn't exist.

imperial_coldfront_plugin/gpfs_client.py
cc-a: It provides additional permissions to the user group. Still a placeholder value that we'll likely change later.

PR #198, opened 2025-03-28
Ldap consistency check

### LINKED ISSUE #169: AD - compare state of DB and AD
Our integration aims to keep the data in AD in sync with the Coldfront database. This could likely go wrong in a number of ways so the idea is to have a scheduled task that looks at the data in both and checks they are consistent. If there are discrepancies these get flagged to system admins for manual review.

On the Coldfront side the data is:
- For each  `Allocation` instance that matches the query `Allocation.objects.filter(resources__name="RDF Project Storage Space", status__name="Active", allocationattribute__allocation_attribute_type__name="RDF Project ID")` there should be an equivalent AD group.
- For each of the above allocation instances the members in the AD group should match to the `AllocationUser` instances that match the query `AllocationUser.objects.filter(allocation=allocation, status__name="Active")`. These should match exactly no additional entries in either.

Sahil590:
# Description
Could not really test this, just went off the documentation and the previous examples. 

**The rest of the description is copilot generated**
This pull request introduces a new feature to check LDAP consistency with the database and includes various supporting changes across multiple files. The most important changes include the addition of the `check_ldap_consistency` function, the scheduling of this consistency check, and the corresponding tests to ensure its functionality.

### New Feature: LDAP Consistency Check

* [`imperial_coldfront_plugin/ldap.py`](diffhunk://#diff-507d719cd54ef71bd41169d94c098782d075726bd1a6285a56237039e298181dR137-R277): Added the `check_ldap_consistency` function, which checks the consistency of LDAP groups with the database and logs any discrepancies. It also includes `_send_discrepancy_notification` to notify admins of any issues found.

### Supporting Changes

* [`imperial_coldfront_plugin/ldap.py`](diffhunk://#diff-507d719cd54ef71bd41169d94c098782d075726bd1a6285a56237039e298181dR3-R8): Added necessary imports such as `logging`, `mail_admins`, and models from `coldfront.core.allocation`.
* [`imperial_coldfront_plugin/tasks.py`](diffhunk://#diff-db0fb86fa77d71ad0600028e218510131bbfc41ca1e231ca7cc31c15b778d094R40-R50): Introduced the `run_consistency_check` function to schedule the LDAP consistency check to run daily.
* [`imperial_coldfront_plugin/tasks.py`](diffhunk://#diff-db0fb86fa77d71ad0600028e218510131bbfc41ca1e231ca7cc31c15b778d094L7-R7): Added the `schedule` import from `django_q.tasks`.

Fixes #169 

## Type of change

## Key checklist

## Further checks


imperial_coldfront_plugin/ldap.py
```diff
@@ -130,6 +134,147 @@ def _ldap_remove_member_from_group(
...
+                message.append(
+                    f"- Allocation ID {err['allocation_id']}: {err['error']}"
+                )
+            else:
+                message.append(f"- {err['error']}")
+
+    if discrepancies.get("ldap_search_errors"):
+        message.append("\nLDAP Search Errors:")
+        for err in discrepancies.get("ldap_search_errors", []):
+            message.append(f"- Group {err.get('group_id')}: {err.get('error')}")
```
cc-a, line 268: Don't see anywhere this is populated above.

tests/test_ldap.py
```diff
@@ -82,3 +83,124 @@ def test_ldap_remove_member_from_group_wrong_error_code(ldap_connection_mock):
...
+            return (True, None, [{"attributes": {"cn": [group_id]}}], None)
+        return (True, None, [{"dn": username}], None)
+
+    ldap_connection_mock().search.side_effect = search_side_effect
+
+    mock_send = mocker.patch(
+        "imperial_coldfront_plugin.ldap._send_discrepancy_notification"
+    )
+
+    mocker.patch("imperial_coldfront_plugin.ldap.settings.LDAP_ENABLED", True)
```
cc-a, line 121: This shouldn't be required as the setting is overriden in `conftest.py`.

tests/test_ldap.py
```diff
@@ -82,3 +83,124 @@ def test_ldap_remove_member_from_group_wrong_error_code(ldap_connection_mock):
...
+
+
+def test_check_ldap_consistency_success(
+    ldap_connection_mock, rdf_allocation, allocation_user, mocker
+):
+    """Test check_ldap_consistency when everything is in sync."""
+    group_id = rdf_allocation.allocationattribute_set.get(
+        allocation_attribute_type__name="RDF Project ID"
+    ).value
+    username = allocation_user.user.username
```
cc-a, line 95: This is available as the fixture `rdf_allocation_project_id`

tests/test_ldap.py
```diff
@@ -82,3 +83,124 @@ def test_ldap_remove_member_from_group_wrong_error_code(ldap_connection_mock):
...
+                    {
+                        "attributes": {
+                            "member": [f"cn={username},ou=Users,dc=example,dc=com"]
+                        }
+                    }
+                ],
+                None,
+            )
+        elif "(cn=rdf-*)" in filter_query:
+            return (True, None, [{"attributes": {"cn": [group_id]}}], None)
```
cc-a, line 112: What does this logic cover?

imperial_coldfront_plugin/tasks.py
```diff
@@ -37,3 +37,14 @@ def wrapped(*args, **kwargs):
...
+
+def run_consistency_check():
+    """Run the LDAP consistency check task."""
+    # Schedule the task to run daily
+    schedule(
+        "imperial_coldfront_plugin.tasks.check_ldap_consistency",
+        schedule_type="D",
+        repeats=-1,
+        next_run=datetime.datetime.now(),
+    )
```
cc-a, line 50: This function is never called so the task won't actually get scheduled. Instead we use a data migration to make sure it is always present in the database see the expiry notification email task as an example https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/main/imperial_coldfront_plugin/migrations/0009_schedule_expiry_notification.py

cc-a (changes requested):
Thanks @Sahil590. Needs another pass I think.

imperial_coldfront_plugin/ldap.py
```diff
@@ -130,6 +134,147 @@ def _ldap_remove_member_from_group(
             )
 
 
+def check_ldap_consistency():
+    """Check the consistency of LDAP groups with the database."""
+    logger = logging.getLogger(__name__)
+    discrepancies = {
+        "missing_groups": [],
+        "extra_groups": [],
```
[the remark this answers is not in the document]
Sahil590: I used some of the copilot code in this and had some of these checks forgot to remove this.

tests/test_ldap.py
Sahil590: I will have to relook at the tests I had used copilot for them so they are not the best

imperial_coldfront_plugin/ldap.py
Sahil590: This was not needed

imperial_coldfront_plugin/tasks.py
```diff
@@ -37,3 +37,14 @@ def wrapped(*args, **kwargs):
...
+
+
+def run_consistency_check():
+    """Run the LDAP consistency check task."""
+    # Schedule the task to run daily
+    schedule(
+        "imperial_coldfront_plugin.tasks.check_ldap_consistency",
+        schedule_type="D",
+        repeats=-1,
+        next_run=datetime.datetime.now(),
```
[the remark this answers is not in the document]
Sahil590: datetime module is already added at the top of the file

imperial_coldfront_plugin/ldap.py
```diff
@@ -130,6 +134,63 @@ def _ldap_remove_member_from_group(
             )
 
 
+def check_ldap_consistency():
+    """Check the consistency of LDAP groups with the database."""
+    discrepancies = {
+        "membership_discrepancies": [],
+    }
```
cc-a, line 141: Doesn't seem much point to have a dictionary with a single key here.

imperial_coldfront_plugin/migrations/0011_auto_20250401_1022.py
```diff
@@ -0,0 +1,27 @@
...
+def run_consistency_check(apps, schema_editor):
+    """Schedule a task to run the consistency check."""
+    Schedule = apps.get_model("django_q", "Schedule")
+    schedule(
+        "imperial_coldfront_plugin.tasks.check_allocation_user_consistency",
+        schedule_type="I",
+        minutes=60,
+        repeats=-1,
+        next_run=datetime.datetime.now(),
+    )
```
cc-a, line 17: Not checked the below but you need to use the Schedule model.
```suggestion
    Schedule.objects.update_or_create(
        name="LDAP consistency check",
        func="imperial_coldfront_plugin.tasks.check_allocation_user_consistency",
        schedule_type="D",
    )
```

tests/test_ldap.py
```diff
@@ -82,3 +85,168 @@ def test_ldap_remove_member_from_group_wrong_error_code(ldap_connection_mock):
     )
     with pytest.raises(LDAPGroupModifyError):
         _ldap_remove_member_from_group(GROUP_NAME, MEMBER_USERNAME, allow_missing=True)
+
+
+@patch("imperial_coldfront_plugin.ldap._send_discrepancy_notification")
```
cc-a, line 90: Do we need to mock this out? Without it we could used the `mailoutbox` fixture to check the email is/isn't being sent.

cc-a (changes requested):
Thanks @Sahil590. Looking better. Just a handful of comments to follow up on.

tests/test_ldap.py
Sahil590: i chose to mock is because of the way the mail_admins work in Django was having trouble using the mailoutbox

cc-a: Interesting, what was the issue? Probably not worth fixing if its a pain but I'm curious.

Sahil590: I was having issues with it actually sending the email when running the tests, when checking the `mailoutbox`. I thought it was because of not specifying the admins in the mock settings, but that did not fix it. After that I was not sure what why so just decided to use this patch instead.

cc-a: Ok, let's leave it as is. I'll make a note to test it fires properly in production.

cc-a (approved):

PR #200, opened 2025-04-01
Add DART id to `RDFAllocation`

### LINKED ISSUE #196: Add DART id's
As a matter of policy all rdf allocations are expected to have an associated [DART](https://www.imperial.ac.uk/admin-services/secretariat/policies-and-guidance/data-assessments/) entry. This should be requested as a field on the `RDFAllocation` form and associated to the allocation via a new allocation attribute type. The data should be non-editable for end users.

SaranjeetKaur:
# Description

This pull request adds a `dart_id` field to the `RDFAllocationForm` class which is not  editable by end users.

Fixes #196 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -546,6 +546,7 @@ def add_rdf_storage_allocation(request):
             storage_size_gb = form.cleaned_data["size"]
             faculty = form.cleaned_data["faculty"]
             department = form.cleaned_data["department"]
+            dart_id = form.cleaned_data["dart_id"]
 
             rdf_id_attribute_type = AllocationAttributeType.objects.get(
                 name="RDF Project ID"
@@ -556,6 +557,13 @@ def add_rdf_storage_allocation(request):
                 value=project_id,
             ):
                 raise ValueError("RDF project with ID already exists.")
+
+            dart_id_attribute_type = AllocationAttributeType.objects.get(name="DART ID")
+            if AllocationAttribute.objects.filter(
+                allocation_attribute_type=dart_id_attribute_type,
+                value=dart_id,
+            ):
+                raise ValueError("DART ID already exists.")
 
             rdf_resource = Resource.objects.get(name="RDF Project Storage Space")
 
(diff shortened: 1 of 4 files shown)


SaranjeetKaur:
(I didn't know Copilot could review as well, so tried it on this one)

cc-a (changes requested):
Thanks @SaranjeetKaur, there's another part to this however. The RDFAllocationForm collects the DART id (the field should be enabled here) from an admin. You then need to store the value from the form as an `AllocationAttribute`. You can see how this is done for the `RDF Project ID` attribute type in the `add_rdf_storage_allocation` view function.

imperial_coldfront_plugin/views.py
```diff
@@ -557,6 +558,15 @@ def add_rdf_storage_allocation(request):
             ):
                 raise ValueError("RDF project with ID already exists.")
 
+            dart_id_attribute_type = AllocationAttributeType.objects.get(
+                name="DART ID", is_changeable=False
```
SaranjeetKaur, line 562: @cc-a is it ok to name it as `"DART ID"`?

SaranjeetKaur:
Can't seem to figure out how to fix the failing tests! Will open for a review.

imperial_coldfront_plugin/views.py
```diff
@@ -585,6 +586,15 @@ def add_rdf_storage_allocation(request):
                 value=project_id,
             )
 
+            dart_id_attribute_type = AllocationAttributeType.objects.get(
+                name="DART ID", is_changeable=False
```
SaranjeetKaur, line 590: Question: Did we want to have the `is_private=True` set here? Would imply that the end users won't be to see the attribute?

```
If you do not want the PI and allocation users to see a particular attribute, check the "Is private" checkbox. System administrators will be able to view these attributes on the allocation detail page but they'll be marked with a lock icon. Any others viewing the allocation will not see them at all.
```

cc-a: No, shouldn't be private.

cc-a (approved):
Great. Looks good.

PR #206, opened 2025-04-09
Retrieving quota usage

### LINKED ISSUE #195: Quota usage reporting
~~We want to display to users via Coldfront how much of their storage allocation quota they are using. This is well supported by Coldfront via the `AllocationAttributeUsage` model. We need to:~~

~~- extend the `add_rdf_storage_allocation` view function to create a AllocationAttributeUsage against the "Storage Quota" attribute.~~
~~- add support for the retrieving quota usage data to the GPFS API client - there's an endpoint [for quotas on individual filesets](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamefilesetsfilesetnamequotas-get) or it may be possible to use the [filesystem level endpoint](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamequotas-get).~~
~~- create a scheduled task that will get the quota use data and update the allocation usage models.~~

Truncating the Coldfront related aspects of this work for now. The only point we can look at:
- add support for the retrieving quota usage data to the GPFS API client - there's an endpoint [for quotas on individual filesets](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamefilesetsfilesetnamequotas-get) or it may be possible to use the [filesystem level endpoint](https://www.ibm.com/docs/en/storage-scale/5.2.2?topic=endpoints-filesystemsfilesystemnamequotas-get).

SaranjeetKaur:
# Description

Retrieve the quota usage of a fileset in the `gpfs_client.py` file.

Fixes #195 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -255,7 +255,6 @@ class GPFSClient(Consumer):
                 return task_data
             raise
 
-    @check_job_status
     @json
     @get("filesystems/{filesystemName}/filesets/{filesetName}/quotas")
     def _retrieve_quota_usage(
@@ -274,8 +273,8 @@ class GPFSClient(Consumer):
         """Method (public) to retrieve the quota usage of a fileset.
 
         Args:
-            filesystem_name: Name of the filesystem where the quota will be set.
-            fileset_name: Name of the fileset to set the quota.
+            filesystem_name: Name of the filesystem to retrieve the quota usage from.
+            fileset_name: Name of the fileset to retrieve the quota usage from.
 
         Returns:
             The response after successfully retrieving the quota usage.
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -255,6 +255,35 @@ class GPFSClient(Consumer):
                 return task_data
             raise
 
+    @check_job_status
+    @json
+    @get("filesystems/{filesystemName}/filesets/{filesetName}/quotas")
+    def _retrieve_quota_usage(
+        self,
+        filesystemName: str,
+        filesetName: str,
+    ) -> requests.Response:
+        """Method (private) to retrieve the quota usage of a fileset."""
+        pass
+
+    def retrieve_quota_usage(
+        self,
+        filesystem_name: str,
+        fileset_name: str,
+    ) -> requests.Response:
+        """Method (public) to retrieve the quota usage of a fileset.
+
+        Args:
+            filesystem_name: Name of the filesystem where the quota will be set.
+            fileset_name: Name of the fileset to set the quota.
+
+        Returns:
+            The response after successfully retrieving the quota usage.
+        """
+        return self._retrieve_quota_usage(
+            filesystemName=filesystem_name, filesetName=fileset_name
+        )
+
 
 class FilesetCreationError(Exception):
     """Raised when a problem is encountered when creating a fileset."""
--- a/imperial_coldfront_plugin/gpfs_client.py
+++ b/imperial_coldfront_plugin/gpfs_client.py
@@ -277,11 +277,27 @@ class GPFSClient(Consumer):
             fileset_name: Name of the fileset to retrieve the quota usage from.
 
         Returns:
-            The response after successfully retrieving the quota usage.
+            The block and files usage values.
         """
-        return self._retrieve_quota_usage(
+        response = self._retrieve_quota_usage(
             filesystemName=filesystem_name, filesetName=fileset_name
         )
+        data = response.json()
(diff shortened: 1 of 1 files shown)


imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -255,6 +255,35 @@ def create_fileset_directory(
                 return task_data
             raise
 
+    @check_job_status
```
cc-a, line 258: Not needed and causes an error as GET endpoints just return data and don't trigger the GPFS server to run a background job.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -255,6 +255,35 @@ def create_fileset_directory(
...
+        Args:
+            filesystem_name: Name of the filesystem where the quota will be set.
+            fileset_name: Name of the fileset to set the quota.
+
+        Returns:
+            The response after successfully retrieving the quota usage.
+        """
+        return self._retrieve_quota_usage(
+            filesystemName=filesystem_name, filesetName=fileset_name
+        )
```
cc-a, line 285: It would be nice to do some processing of the return data so that we only get the block and files quota values.

cc-a (changes requested):

SaranjeetKaur:
Processed the return data to retrieve the following:
```
block_usage = data["quotas"][0]["blockUsage"]  # denotes the "Usage": Current capacity quota usage.
block_quota = data["quotas"][0]["blockQuota"]  # denotes the"Soft limit": The soft limit set for the fileset, user, or user group.
block_limit = data["quotas"][0]["blockLimit"]  # denotes the "Hard limit": The hard limit set for the capacity quota usage. A grace period starts when the hard limit is reached.
files_usage = data["quotas"][0]["filesUsage"]  # denotes the "Number of files in usage": Number of inodes.
files_quota = data["quotas"][0]["filesQuota"]  # denotes the "Soft limit": The soft limit set for the inode quota.
files_limit = data["quotas"][0]["filesLimit"]  # denotes the "Hard limit": The hard limit set for the inode quota.
```

I've tried to retrieve more than what was asked for, block and files usage, respectively - it is only for trial purpose. Can remove the extra bits or leave them there, as per feedback on the PR.

Steps to test these changes:

Launch the coldfront shell using:

```bash
$ docker compose exec app coldfront shell
```

```output
Python 3.11.12 (main, Apr  9 2025, 18:23:23) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
```

```python
>>> from imperial_coldfront_plugin.gpfs_client import GPFSClient
>>> client = GPFSClient()
>>> response = client.retrieve_quota_usage("gpfs", "rdf-000002")    # using your username and password on https://gpfs-dev-1.hpc.ic.ac.uk, find an existing Fileset (here, `rdf-000002`) whose quota usage is to be retrieved
```

If `print(data)` and `print(retrieved_data)` is used within the functions and the above code is run in the shell, it should give the following output:

```output
/usr/local/lib/python3.11/site-packages/urllib3/connectionpool.py:1103: InsecureRequestWarning: Unverified HTTPS request is being made to host 'gpfs-dev-1.hpc.ic.ac.uk'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings     
  warnings.warn(
{'quotas': [{'blockGrace': 'none', 
                  'blockInDoubt': 0, 
                  'blockLimit': 0, 
                  'blockQuota': 0, 
                  'blockUsage': 0, 
                  'filesGrace': 'none', 
                  'filesInDoubt': 0, 
                  'filesLimit': 0, 
                  'filesQuota': 0, 
                  'filesUsage': 2, 
                  'filesetName': 'rdf-000002', 
                  'filesystemName': 'gpfs', 
                  'isDefaultQuota': False, 
                  'objectId': 0, 
                  'objectName': 'root', 
                  'quotaId': 191, 
                  'quotaType': 'GRP'}, 
                  {'blocpe': 'USR'}], 
 'status': {'code': 200, 
               'message': 'The request finished successfully.'}}

{'block_usage': 0, 
 'block_quota': 0,
 'block_limit': 0, 
 'files_usage': 2, 
 'files_quota': 0, 
 'files_limit': 0} 
```

cc-a (commented):
Thanks @SaranjeetKaur. Working for me now. I think we should just return the usage values for the blocks and files.

SaranjeetKaur:
@cc-a I've removed the other values now

cc-a (approved):
Great. Thanks @SaranjeetKaur

PR #216, opened 2025-05-07
:recycle: Streamline allocation creation

### LINKED ISSUE #213: Update add_rdf_storage_allocation to possibly also create a Project
At the moment we populate the `project` field of `RDFAllocationForm` with a list of existing projects within which to create the storage allocation. This basically assumes that the project owner has already logged in and created the project themselves. In order to provide a simpler setup process we're going to support creating a user for the group owner and a project before creating and adding the storage allocation.

To do:
- Replace `RDFAllocationForm.project` with a `username` CharField.
- In `add_rdf_storage_allocation` update the form data processing:
  - To try and lookup a user in the database using the username. If it a user doesn't exist, create one using data from `MicrosoftGraphClient.user_profile` (created user should have the username, email, first and last names set as a minimum).
  - Check if that user has an existing project. If not then create one.
  - Add the storage allocation to the retrieved/created project.

dalonsoa:
# Description

Updates the storage allocation process to also create a project - and a user, if needed -, so things are simpler for the end user. 

Fixes #213 

## Type of change

- [x] Optimization (non-breaking, back-end change that speeds up the code)

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/views.py
+++ b/imperial_coldfront_plugin/views.py
@@ -533,18 +533,7 @@ def add_rdf_storage_allocation(request):
 
             # We create a new user and an associated project, if they don't exist.
             user = get_or_create_user(form.cleaned_data["username"])
-            project, project_created = Project.objects.get_or_create(
-                pi=user,
-                title=f"{user.get_full_name()}'s Research Group",
-                status=ProjectStatusChoice.objects.get(name="Active"),
-            )
-            if project_created:
-                ProjectUser.objects.create(
-                    user=user,
-                    project=project,
-                    role=ProjectUserRoleChoice.objects.get(name="Manager"),
-                    status=ProjectUserStatusChoice.objects.get(name="Active"),
-                )
+            project = get_or_create_project(user)
 
             rdf_allocation = Allocation.objects.create(
                 project=project,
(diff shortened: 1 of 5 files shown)


dalonsoa:
@cc-a could you have a look at this? If it makes sense, I'll work on the tests. My main unknown is the creation of a new project. I'm not sure if the input arguments are the ones that shall be use. In particular, `title` and `status` are mandatory, and I don't know if I've made reasonable choices there.

cc-a (commented):
Thanks @dalonsoa. Looking good, just a couple of small tweaks suggested.

cc-a:
> It seems that the default field of science is not available, which makes it quite useless as default field of science. @cc-a o you have any suggestion about how to proceed? Maybe getting or creating the default field when creating a project?

Odd,... Why does it complain in the test teardown? They've done something really gross with a [hardcoded pk value](https://github.com/ubccr/coldfront/blob/c766ee41a36c4e9a8f38740489624b2133d21f7b/coldfront/core/field_of_science/models.py#L22) for some reason. Hopefully we can just handle it in the same way as we have some previously prerequisite models via creating it in the `rdf_allocation_dependencies` fixture.

dalonsoa:
Quick and dirty solution in the test that is causing trouble. Not sure why that one causes trouble, though...

imperial_coldfront_plugin/views.py
```diff
@@ -631,3 +641,50 @@ def task_stat_view(request, group: str):
...
+        user: The user object that will own a project.
+
+    Return:
+        The project, already existing or newly created.
+    """
+    project, project_created = Project.objects.get_or_create(
+        pi=user,
+        title=f"{user.get_full_name()}'s Research Group",
+        status=ProjectStatusChoice.objects.get_or_create(name="Active")[0],
+    )
```
cc-a, line 682: I'd just make a tiny tweak here to avoid creating a second group in the event that a project title gets edited or we decide to change the naming pattern later.
```suggestion
    project, project_created = Project.objects.get_or_create(
        pi=user,
        defaults=dict(
            title=f"{user.get_full_name()}'s Research Group",
            status=ProjectStatusChoice.objects.get_or_create(name="Active")[0],
        )
    )
```

cc-a (approved):
Thanks @dalonsoa. Looks great. I have one small tweak.

imperial_coldfront_plugin/views.py
dalonsoa: Well spotted!

PR #217, opened 2025-05-07
Rename RDF resource to "RDF Active"

cc-a:
# Description

For simplicity and consistency this PR renames the RDF storage resource type from "RDF Project Storage Space" to "RDF Active". This should be clearer to users as "RDF Active" is the name of the actual service they're consuming. Note this is a bit naughty to modify the existing data migration. You'll need to wipe the existing database (`docker compose down -v`) to get a consistent state after this PR is merged.

## Type of change

- [X] Breaking change (whatever its nature)

## Key checklist

## Further checks


dalonsoa (approved):
Well, naughty or not, now it is the time to do this. No one will notice this was not `RDF Active` from the start.

PR #218, opened 2025-05-07
Add files quota

cc-a:
# Description

In addition to the used storage quota we need to track the number of files in use in each storage allocation. It's a pretty straight extension of what we use for storage quotas. More naughty fiddling with data migrations I'm afraid.

Fixes # (issue)

## Type of change

- [X] Breaking change (whatever its nature)

## Key checklist

## Further checks


imperial_coldfront_plugin/views.py
```diff
@@ -531,17 +532,33 @@ def add_rdf_storage_allocation(request):
...
+                allocation_attribute=quota_attribute, value=0
+            )
+
+            files_quota_attribute_type = AllocationAttributeType.objects.get(
+                name="Files Quota"
+            )
+            files_attribute = AllocationAttribute.objects.create(
+                allocation_attribute_type=files_quota_attribute_type,
+                allocation=rdf_allocation,
+                value=settings.GPFS_FILES_QUOTA,
```
dalonsoa, line 557: Was this `GPFS_FILES_QUOTA` already defined in settings? For what was it used?

dalonsoa (approved):
This looks OK, however I'm a bit confused. Why are you editing past migrations instead of just creating new ones with the new functionality, which seems more canonical?

imperial_coldfront_plugin/views.py
cc-a: It's used [here](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/a92b77fad34f9168be9461876d26ada1cfa3a1cd/imperial_coldfront_plugin/views.py#L599-L611) when creating a fileset and setting the initial quota values. It provides the maximum number of files that can be created in a fileset. It's expected to be a fixed value for all filesets.

cc-a:
> This looks OK, however I'm a bit confused. Why are you editing past migrations instead of just creating new ones with the new functionality, which seems more canonical?

Yeah, fair. Was trying to cut a corner and as noted previously all the migrations need rationalising. I've moved things into a new migration for now.

PR #222, opened 2025-05-08
Add GIDs

### LINKED ISSUE #212: GIDs
- Add a private AllocationAttributeType of "GID" via data migration.
- Add a setting that defines a set of non-contiguous number ranges from which GID's should be picked. 
- Create a function to select a value for a new GID that should pick the lowest available value from the configured ranges - it will need to check what values have already been assigned by checking for all existing allocation attributes of type "GID".
- Update `_ldap_create_group` to take a gid value as an argument and to provide it as the attribute `gidNumber` when calling `conn.add`.
- Update the `add_rdf_storage_allocation` view function to add a GID attribute to the allocation and to pass it to the ldap function.

SaranjeetKaur:
# Description

_Please include a summary of the change and which issue is fixed (if any). Please also
include relevant motivation and context. List any dependencies that are required for
this change._

Fixes #212 

## Type of change

## Key checklist

## Further checks


imperial_coldfront_plugin/settings.py
```diff
@@ -57,3 +57,8 @@
 """The organisational unit containing RDF access groups."""
 
 LDAP_ENABLED = bool(LDAP_USERNAME and LDAP_PASSWORD and LDAP_URI)
+
+GID_RANGES = [
+    range(1000000, 1999999),
+    range(2000000, 2999999),
```
SaranjeetKaur, line 63: An example range - not sure what range to use here.

SaranjeetKaur:
The tests aren't passing yet. But better to pause and ask if I am going in the right direction.

imperial_coldfront_plugin/gid.py
```diff
@@ -0,0 +1,40 @@
...
+        ValueError: If no available GID is found in the configured ranges.
+    """
+    existing_gids = Allocation.objects.filter(
+        allocationattribute__allocation_attribute_type__name="GID"
+    ).values_list("allocationattribute__value", flat=True)
+
+    # Get the maximum GID value already assigned
+    max_gid = existing_gids.aggregate(Max("allocationattribute__value"))[
+        "allocationattribute__value__max"
+    ]
```
cc-a, line 28: This looks like it should work but it doesn't because of the funny way they handled how attribute types work where all values are actually internally stored as strings. This is a bit crude but should do the trick (assuming the attribute type is changed to `Int`).
 ```suggestion
    existing_gids = AllocationAttribute.objects.filter(
        allocation_attribute_type__name="GID"
    )

    # Get the maximum GID value already assigned
    max_gid = max(eg.typed_value() for eg in existing_gids)
```

File: imperial_coldfront_plugin/views.py [hunk 3]
```diff
@@ -585,6 +586,17 @@ def add_rdf_storage_allocation(request):
...
+            gid = get_new_gid()
+            AllocationAttribute.objects.create(
+                allocation_attribute_type=gid_attribute_type,
+                allocation=rdf_allocation,
+                value=gid,
+            )
+            AllocationAttributeUsage.objects.create(
+                allocation_attribute=gid_attribute_type, value=0
+            )
+
```
cc-a, line 599: Don't forget to pass the gid value below when `_ldap_create_group` is added to the chain below.

cc-a (commented):
Hey @SaranjeetKaur. Looking good. Some tweaks needed due to some of the unintuitive Coldfront models.

cc-a:
@SaranjeetKaur I've taken the liberty of polishing this up as you're not feeling well. I'll leave an issue for testing in the next sprint.

cc-a (approved):

PR #223, opened 2025-05-08
Strip out group related functionality

cc-a:
# Description

Removes the no longer needed functionality after our change of direction. May be restored later but I'd like a clean initial deployment that we can revise later. I'm going to take a pass at simplify all of the migrations down in a subsequent PR.

Fixes #

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
All of these effort... :( These are weeks worth of work wiped out. Well, I guess that the beauty of version control is that we can always come back to it. Before merging, I'd suggest you tag the main branch so we can find the point in history where we had groups in case we need them in the future. 

Can you remind me why we don't need groups anymore?

PR #224, opened 2025-05-08
More sophisticated treatment of Dart ID's

cc-a:
# Description

Validates and prevents duplicates of Dart ID values set as AllocationAttributes. Also provides a dedicated route for users/superusers to add additional Dart ID's to an allocation. Also updates the `DART ID` AllocationAttributeType to not be private.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/dart.py
+++ b/imperial_coldfront_plugin/dart.py
@@ -24,7 +24,7 @@ def validate_dart_id(dart_id: str):
         raise DartIDValidationError("Dart ID already assigned to an allocation")
 
 
-def create_dart_id_attribute(dart_id: int, allocation: Allocation):
+def create_dart_id_attribute(dart_id: str, allocation: Allocation):
     """Create an AllocationAttribute for a Dart ID."""
     validate_dart_id(dart_id)
     dart_id_attribute_type = AllocationAttributeType.objects.get(
--- a/imperial_coldfront_plugin/dart.py
+++ b/imperial_coldfront_plugin/dart.py
@@ -1,0 +1,37 @@
+"""Module for creation and validation of Dart ID AllocationAttribute's."""
+
+from coldfront.core.allocation.models import (
+    Allocation,
+    AllocationAttribute,
+    AllocationAttributeType,
+)
+
+
+class DartIDValidationError(Exception):
+    """Error when a Dart ID does not meet validation criteria."""
+
+
+def validate_dart_id(dart_id: str):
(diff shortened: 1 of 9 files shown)


imperial_coldfront_plugin/forms.py
```diff
@@ -120,6 +122,15 @@ class RDFAllocationForm(forms.Form):
         disabled=False,
     )
 
+    def clean_dart_id(self) -> str:
```
dalonsoa, line 125: This is one of those methods that are magically called by Django, based on the variable name, right? I'm asking because I don't see it being used anywhere - at least not explicitly.

imperial_coldfront_plugin/dart.py
```diff
@@ -0,0 +1,37 @@
...
+    """Create an AllocationAttribute for a Dart ID."""
+    validate_dart_id(dart_id)
+    dart_id_attribute_type = AllocationAttributeType.objects.get(
+        name="DART ID", is_changeable=False
+    )
+    return AllocationAttribute.objects.create(
+        allocation_attribute_type=dart_id_attribute_type,
+        allocation=allocation,
+        value=dart_id,
+    )
```
dalonsoa, line 37: I understand that this is just the way Coldfront do things, but do you know what's the rationale for this approach rather than just having a `DartID` model with `Allocation` as foreign key and a value? I'm sure there's a good reason for this - I would say that facilitating having arbitrary number of attributes without an explosion of models for each of them - but just checking to make sure I understand this right.

dalonsoa (approved):
It looks good to me. We are not checking if that Dart ID actually exist in Dart, right? In that sense, people could just enter a random integer and, assuming it is not taken, it will be OK from the point of view of this system.

imperial_coldfront_plugin/dart.py
cc-a: So Coldfront has implemented a fairly complex data model that allows for attaching arbitrary metadata to allocations via these atributes. I think the rationale is supposed to be that you can create attribute types etc. all via the admin interface without the need to write code. I guess that's kind of ok if you just want to manually tag arbitrary metadata but it's not a great approach IMO if you want to build more sophisticated behaviour around those attributes as you just end up having to handle these things via data migrations and it's all rather cumbersome.

imperial_coldfront_plugin/forms.py
cc-a: Yeah that's right. Form magic.

PR #225, opened 2025-05-08
Add a scheduled task to update quota usages

cc-a:
# Description

This PR adds:
- a GPFSClient method to retrieve all quota data for all filesets on a filesystem
- a function to update relevant allocation attributes with usage values
- a scheduled task calling the above function hourly

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -293,12 +293,36 @@ def retrieve_quota_usage(
...
+        """Get the quotas for all filesets.
+
+        Arguments:
+            filesystem_name: Name of the filesystem to retrieve the quota usage from.
+        """
+        data = self._retrieve_all_fileset_quotas(filesystem_name).json()
+        return {
+            quota["objectName"]: {
+                "files": quota["filesUsage"],
+                "block_gb": quota["blockUsage"] / 1024**2,
```
dalonsoa, line 320: I would suggest to be consistent with the naming of objects that represent the same things. Above you use `files_usage` and `block_usage_gb`, so here we should use the same names. Or the other way around and change the code above. Either way works fine as long as there is consistency.

imperial_coldfront_plugin/gpfs_client.py
```diff
@@ -355,3 +379,31 @@ def _create_fileset_set_quota(
...
+    """Update the usages of all quota related allocation attributes."""
+    from coldfront.core.allocation.models import Allocation
+
+    client = GPFSClient()
+    usages = client.retrieve_all_fileset_usages(settings.GPFS_FILESYSTEM_NAME)
+
+    # use prefetch_related to reduce number of database operations
+    allocations = Allocation.objects.filter(
+        resources__name="RDF Active"
+    ).prefetch_related("allocationattribute_set")
```
dalonsoa, line 394: Neat! I did not know about this one.

imperial_coldfront_plugin/migrations/0014_auto_20250508_2158.py
```diff
@@ -0,0 +1,21 @@
...
+
+from django.db import migrations
+
+def schedule_update_quota_usages_task(apps, schema_editor):
+    Schedule = apps.get_model("django_q", "Schedule")
+    Schedule.objects.update_or_create(
+        name="Update quota usages",
+        func="imperial_coldfront_plugin.gpfs_client._update_quota_usages_task",
+        schedule_type="H",
+    )
```
dalonsoa, line 11: Are these things typically meant to be added manually to migrations? Seems cumbersome.

dalonsoa (approved):
I'm approving to speed up things, but I think there's some work to do to improve the consistency of the names across the code.

imperial_coldfront_plugin/gpfs_client.py
cc-a: Good point. I've standardised.

imperial_coldfront_plugin/migrations/0014_auto_20250508_2158.py
cc-a: I think it's the best way to handle them as they are database entries. The only alternative would be to just stick it somewhere where it would get created during the app initialisation process but that means you don't have clean options to remove it later.

PR #233, opened 2025-05-29
Update login redirect setting

cc-a:
Minor settings tweak.


SaranjeetKaur (approved):
Redirecting to home after logout? Looks ok

PR #282, opened 2025-12-05
Credits project detail

### LINKED ISSUE #269: Add display of credits to project detail view
Template details to be provided...

Add some content to the project detail view that clearly and prominently displays the current balance. As the project detail view function is a core part of Coldfront this will have to be implemented by a [custom template tag](https://docs.djangoproject.com/en/5.2/howto/custom-template-tags/). See the [get_user_projects](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/1127afdbaca6305d53f9bebfc0ba62b761a57463/imperial_coldfront_plugin/templatetags/projects.py) tag as an example.

Visibility of the above template modifications should be controlled by a feature flag. Adding the setting `CREDITS_ENABLED` is suggested for this. You'll also need to modify the setting [SETTINGS_EXPORT](https://github.com/jkbrzt/django-settings-export?tab=readme-ov-file#usage) to make this available within the template context.

Sahil590:
# Description
I added the new section to the project-detail view to display the credits available to a certain project. This section can be hidden by a setting change `SHOW_CREDIT_BALANCE`. If disabled, it will hide the section. 
I have added the necessary tests for this change, 
For this to fully work, you will need this change to the local settings found in this [PR](https://github.com/ImperialCollegeLondon/coldfront_development_environment/pull/8)

Fixes #269 
Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -93,7 +93,6 @@ def pytest_configure():
             GPFS_API_USERNAME="",
             GPFS_API_PASSWORD="",
             GID_RANGES=[range(1031386, 1031435)],
-            # SHOW_CREDIT_BALANCE=True,
         ),  # override settings loaded by env var for tests
     )
 


imperial_coldfront_plugin/templates/imperial_coldfront_plugin/overrides/project_detail.html
```diff
@@ -23,6 +25,34 @@ <h2 class="text-justify">{{ project.title }}</h2>
...
+        {% get_project_credit_balance project as credit_balance %}
+        <h2 class="mb-0">
+          {% if credit_balance >= 0 %}
+            <span class="badge badge-success badge-lg" style="font-size: 1.5rem; padding: 0.5rem 1rem;">
+              {{ credit_balance|intcomma }} credits
+            </span>
+          {% else %}
+            <span class="badge badge-danger badge-lg" style="font-size: 1.5rem; padding: 0.5rem 1rem;">
+              {{ credit_balance|intcomma }} credits
+            </span>
```
cc-a: Don't need to duplicate so much here as the only thing changing between the two is the badge class.

cc-a (changes requested):

tests/conftest.py
```diff
@@ -89,6 +93,7 @@ def pytest_configure():
             GPFS_API_USERNAME="",
             GPFS_API_PASSWORD="",
             GID_RANGES=[range(1031386, 1031435)],
+            # SHOW_CREDIT_BALANCE=True,
```
cc-a: Can be removed if not used here.

PR #302, opened 2026-01-20
Ldap member rule

### LINKED ISSUE #271: Remove all LDAP group members if an RDF allocation is not active
Currently users are added to the LDAP group for an allocation based solely on their membership, regardless of the status of the allocation. We would like this to only be the case if the allocation is active. To synchronise this with LDAP, we should implement a `post_save` signal for `coldfront.core.allocation.models.Allocation` that removes any current LDAP group members if an `Allocation` is saved that has a non "Active" status. Note that the LDAP group itself should not be deleted.

Sahil590:
# Description

I added a post_save signal for `coldfront.core.allocation.models.Allocation` that removes any current LDAP group members if an Allocation is saved that has a non "Active" status.
Fixes #271 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -347,6 +347,17 @@ def test_remove_allocation_group_members_multiple_users(
     user2 = user_factory()
     user3 = user_factory()
 
+    allocation_user2 = AllocationUser.objects.create(  # noqa F841
+        allocation=rdf_allocation,
+        user=user2,
+        status=allocation_user_active_status,
+    )
+    allocation_user3 = AllocationUser.objects.create(  # noqa F841
+        allocation=rdf_allocation,
+        user=user3,
+        status=allocation_user_active_status,
+    )
+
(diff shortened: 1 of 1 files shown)


imperial_coldfront_plugin/signals.py
```diff
@@ -35,6 +36,18 @@ def _get_shortname_from_allocation(allocation: Allocation) -> str | None:
         return None
 
 
+def _remove_allocation_group_members(group_id: str) -> None:
+    """Background task: remove all members from an LDAP group."""
+    members_by_group = ldap_group_member_search(group_id)
```
cc-a: We shouldn't need to hit LDAP here as we can get the current users from the database. Also this can go in `tasks.py`.

Sahil590:
@cc-a the `check_ldap_consistency` already filters for the allocation status in `status_name`

PR #308, opened 2026-01-29
Use Beautifulsoup for beautiful HTML tests

### LINKED ISSUE #287: Use BeautifulSoup for html tests
Currently we have a number of tests that do some rather crude checks on the html rendered by our templates. These tests are basic and usually focus on the presence of strings somewhere within the document. [BeautifulSoup][https://www.crummy.com/software/BeautifulSoup/bs4/doc/] is an HTML parsing library that can be used to construct more sophisticated queries of html documents that are well suited to tests. As an example of one of our existing tests:

```python
content = response.content.decode("utf-8")
assert "Credit Balance" in content
assert "0 credits" in content
```

using BeautfulSoup we can write more sophisticated checks such as:

```python
def tag_with_text(tag_name, text):
    def match(tag):
        return tag.name == tag_name and text in tag.text
    return match

soup = BeautifulSoup(response.content, "html.parser")
card = soup.find("div", class_="card-body")  # get the first card-body div, further checks are scoped to the card
assert card.find(tag_with_text("h3", "Credit Balance"))  # check for h3 tag containing the text "Credit Balance"
assert card.find(tag_with_text("span", "0 credits"))  # check for span tag containing the text "0 credits"
```

This provides a much more rigorous check of the structure of the html and enables more sophisticated checks.

This issue covers implementing use of BeautifulSoup to replace existing checks of HTML and to identify additional locations that would benefit from checks.

cc-a:
# Description

This PR applies beautiful soup in our test suite to allow us to more precisely test the structure of the HTML generated and hence the logic used in our templates. Generally this means replacing simple string lookups in the HTML with beautifulsoup queries. This means our tests go from making simple checks like "this string is somewhere in the output" to more concrete checks like "this string appears in a div with class X within the first card on the page with id Y".

A couple of incidental changes were made:
- `test_credit_balance_only_visible_to_pi_and_superuser` was not actually using a superuser. Changing this then caused the template to try and lookup some coldfront urls that had to be added to the test `urls.py`.
- Updated the `project` fixture so that the project PI also has a `ProjectUser` entry created so can properly check rendering of the home page when somebody is a member of a group.

Fixes #287 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -648,7 +648,7 @@ class TestProjectDetailView:
         assert not soup.find("div", class_="card", id="credit-balance-card")
 
     def test_credit_balance_only_visible_to_pi_and_superuser(
-        self, rf, project, settings, user_factory, user, superuser
+        self, rf, project, settings, user_factory, superuser
     ):
         """Ensure credit balance section is only rendered for the PI and superusers."""
         settings.SHOW_CREDIT_BALANCE = True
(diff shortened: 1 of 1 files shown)


Sahil590 (approved):
Looks fine to me and easy to implement in future tests.

PR #316, opened 2026-02-10
272 Scheduled task to check GPFS quota consistency

### LINKED ISSUE #272: Implement check_quota_consistency task
We currently have a daily check that the data held in Coldfront and in Active Directory is in sync. We should extend this to GPFS so that we check that data in Coldfront about allocation quotas is correct. 

Coldfront using [Django Q2](https://django-q2.readthedocs.io/en/master/) to handle scheduled task queue. [Scheduled tasks](https://django-q2.readthedocs.io/en/master/schedules.html). Create using a [data migration](https://docs.djangoproject.com/en/6.0/topics/migrations/#data-migrations) see [this example](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/main/imperial_coldfront_plugin/migrations/0002_schedule_ldap_consistency_check.py).

This should be a daily scheduled task that:
- Gets a list of all active rdf storage allocations.
  - e.g. [query here](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/43cd7690ceff31af025648230a71ee32e6f0415b/imperial_coldfront_plugin/tasks.py#L201-L205)
- Gets all fileset quota data from GPFS.
  -Need to update [GPFSClient.retrieve_all_fileset_usages](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/43cd7690ceff31af025648230a71ee32e6f0415b/imperial_coldfront_plugin/gpfs_client.py#L455-L461) to return quota data as well as usage.
  - In the scheduled task use the api client to get the quotas [as here](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/43cd7690ceff31af025648230a71ee32e6f0415b/imperial_coldfront_plugin/tasks.py#L243-L244).
- Checks the two are consistent with one another.
  - In case of discrepancies send an email similar to [here](https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/blob/43cd7690ceff31af025648230a71ee32e6f0415b/imperial_coldfront_plugin/tasks.py#L236C9-L236C38).

sallymatson:
# Description

Compare the allocations to the GPFS filesets to ensure the storage and file quotas are the same. Send an email if not.

Fixes #272

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,11 +15,11 @@ repos:
       - id: pretty-format-yaml
         args: [--autofix, --indent, '2', --offset, '2']
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.36.1
+    rev: 0.36.2
     hooks:
       - id: check-github-workflows
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.0
+    rev: v0.15.1
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]


imperial_coldfront_plugin/emails.py
```diff
@@ -151,3 +151,70 @@ def send_allocation_deletion_notification(
...
+    for discrepancy in discrepancies:
+        shortname = discrepancy["shortname"]
+        message += f"\n- Allocation shortname: {shortname}\n"
+
+        if discrepancy["attribute_storage_quota"]:
+            attribute_storage_quota = discrepancy["attribute_storage_quota"]
+            fileset_storage_quota = discrepancy["fileset_storage_quota"]
+            message += f"""
+                Allocation storage quota of {attribute_storage_quota},
+                 GPFS storage quota of {fileset_storage_quota}.\n"""
```
cc-a: These lines will be heavily indented btw.

imperial_coldfront_plugin/emails.py
```diff
@@ -151,3 +151,70 @@ def send_allocation_deletion_notification(
...
+    """Send notification to admins that fileset was not found.
+
+    Args:
+        shortnames: The allocation shortnames.
+    """
+    if not settings.ADMINS:
+        return
+
+    subject = "Filesets Not Found"
+    message = "The fileset for the following allocation(s) were not found:"
```
cc-a: Could use a similarly expanded message and subject as above.

tests/test_tasks.py
```diff
@@ -605,3 +621,145 @@ def test_check_expiry_notifications_multiple_allocations(
...
+    """Test check_quota_consistency task."""
+    from unittest.mock import patch
+
+    from imperial_coldfront_plugin.tasks import check_quota_consistency
+
+    # Configure allocation and fileset with specific parameters:
+    helper_add_quota_attributes(rdf_allocation, alloc_storage_quota, alloc_files_quota)
+    with patch(
+        "imperial_coldfront_plugin.gpfs_client.GPFSClient.retrieve_all_fileset_usages"
+    ) as mock_usage:
```
cc-a: Tiniest nitpick in the world but can save the with block and indent by using the `mocker` fixture.

cc-a (approved):
Thanks @sallymatson, looks great.

tests/test_tasks.py
sallymatson: Fixed!

PR #329, opened 2026-03-06
Proxy model for RDFAllocation

### LINKED ISSUE #317: Use of proxy models for Allocation and Project
The Coldfront data model is fairly complex and we make heavy use of the `AllocationAttribute` and `ProjectAttribute` models in order to attache the the metadata we need to `Allocation`s and `Project`s. This leads to fairly verbose queries e.g.:
```python
shortname = allocation.allocationattribute_set.get(
       allocation_attribute_type__name="Shortname"
).value
```
peppered throughout the code. We also expect that certain attributes are present and don't always have consistent error checking.

It would be very nice to have a consistent interface that improves code readability and makes sure error handling is done consistently. I believe this can be achieved through the following pattern using [django proxy models](https://docs.djangoproject.com/en/6.0/topics/db/models/#proxy-models). This would look like:

```python
class RDFAllocation(Allocation):
    """Proxy model for RDF Active allocations."""

    class Meta:
        """Meta class for RDFAllocation."""

        proxy = True

    def __init__(self, *args, **kwargs):
        """Ensure RDFAllocation is associated with the 'RDF Active' resource."""
        super().__init__(*args, **kwargs)
        resource = self.get_parent_resource
        if not resource or resource.name != "RDF Active":
            raise ValueError(
                "RDFAllocation must be associated with the 'RDF Active' resource"
            )

    def _get_attribute_value(self, attribute_name: str) -> Any:
        try:
            return self.allocationattribute_set.get(
                allocation_attribute_type__name=attribute_name
            ).typed_value()
        except AllocationAttribute.MultipleObjectsReturned:
            raise ValueError(f"Multiple {attribute_name} attributes found for allocation - {self}")
        except AllocationAttribute.DoesNotExist:
            raise AttributeError(
                f"{attribute_name} attribute not found for this allocation"
            )

    @property
    def shortname(self) -> str:
        """Get the shortname of the allocation."""
        return self._get_attribute_value("Shortname")
```

When working with relevant allocations one simply uses `RDFAllocation` instead of `Allocation`. The database actions it performs are completely equivalent but the Python objects returned have the additional methods.

This issue covers implementing the `RDFAllocation` model with the set of required properties for the expected attributes and testing it. It should also go through the code base to make use the new model where it makes sense. It should also consider implementing an equivalent model for `Project` but this may be of more limited value as we don't access ProjectAttribute values nearly as often.

sallymatson:
# Description

Added the proxy model for RDFAllocation and replaced the attribute queries across the codebase with shorthand. This also allowed me to get rid of some other helper functions and in-line error handling, which is now done in RDFAllocation. 

Fixes #317 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
     hooks:
       - id: check-github-workflows
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.6
+    rev: v0.15.7
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]


imperial_coldfront_plugin/models.py
```diff
@@ -1,16 +1,120 @@
...
+        """Get the shortname attribute of the allocation."""
+        return self._get_attribute("Files Quota")
+
+    @property
+    def files_quota(self) -> int:
+        """Get the shortname of the allocation."""
+        return self.files_quota_attr.typed_value()
+
+
+class RDFProject(Project):
```
cc-a: `RDFProject` is probably a bit of a misnomer actually. Not all projects will have RDF allocations. Maybe `ICLProject` as it reflects the way we've decided to handle projects locally.

imperial_coldfront_plugin/signals.py
```diff
@@ -120,9 +109,14 @@ def sync_ldap_group_membership(
     if not settings.LDAP_ENABLED:
         return
 
-    if (group_id := _get_shortname_from_allocation(instance.allocation)) is None:
+    allocation = instance.allocation
+    shortname = allocation.shortname
```
cc-a: I think we need to default to the assumption that `allocation` may be an `Allocation` or an `RDFAllocation`. I think the tests pass because of the way the test objects are constructed but in general in those could be any allocation not just an RDF one.

sallymatson:
@cc-a I hope I caught all the things you mentioned... I don't feel 100% comfortable vetting because (as mentioned in my q up there ^) I don't quite get when to use `Allocation`. But hopefully I've hit them all... let me know!!

cc-a (changes requested):
Nearly there! Just a few tweaks.

PR #365, opened 2026-03-31
Add consistency check for hx2 allocations and ldap

### LINKED ISSUE #339: Add consistency check task for HX2 UserAllocations and LDAP group membership
This should follow the established pattern for the check on RDF Active group memberships but applied to hx2 allocations and groups instead.

sallymatson:
# Description

Based on the logic for `check_ldap_consistency`.  I added a specification for the email notification when it comes from RDF vs HX2, because otherwise the admin could recieve two emails at the same time with the same subject which I think would be cause for confusion / assuming it is duplicate instead of two separate emails. 

Not sure if instead they should just be linked into the same email... i.e. just add the HX2 search for discrepancies into the same check. If you prefer that @cc-a just let me know and I'll change it! 

I've also organised all the tests in `test_tasks.py` by what function they are testing. This should be siloed into the first commit to make it easier to see what is to do with the actual logic of this PR! 

Fixes #339 

## Type of change

## Key checklist

## Further checks


imperial_coldfront_plugin/emails.py
```diff
@@ -44,7 +50,7 @@ def send_discrepancy_notification(discrepancies: list[Discrepancy]) -> None:
                 message += f"    - {member}\n"
 
     mail_admins(
-        subject="LDAP Consistency Check - Discrepancies Found",
+        subject=f"LDAP Consistency Check - Discrepancies Found in {source}",
```
cc-a:
```suggestion
        subject=f"LDAP Consistency Check - Discrepancies found for {source} allocations",
```

imperial_coldfront_plugin/tasks.py
```diff
@@ -220,6 +220,49 @@ def check_ldap_consistency() -> list[Discrepancy]:
...
+    for allocation in allocations:
+        group_name = allocation.ldap_shortname
+
+        active_users = AllocationUser.objects.filter(
+            allocation=allocation, status__name="Active"
+        )
+        expected_usernames = [au.user.username for au in active_users]
+
+        actual_members = ldap_groups.get(group_name, [])
+        missing_members = set(expected_usernames) - set(actual_members)
```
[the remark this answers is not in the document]
cc-a: I'm fine with this. Any RDFAllocation missing a shortname should be flagged as soon as possible for admin attention. It shouldn't really ever come up. I don't think it's worth the effort to collate the errors and raise later.

tests/test_emails.py
```diff
@@ -85,3 +86,42 @@ def test_send_fileset_not_found_notification():
...
+def test_send_discrepancy_notification_rdf():
+    """Test that the discrepancy email references RDF by default."""
+    discrepancies = [
+        {
+            "project_name": "Test Project",
+            "group_name": "rdfdev-testgroup",
+            "missing_members": ["alice"],
+            "extra_members": [],
+        }
+    ]
```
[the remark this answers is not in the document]
cc-a: Actually it looks like the `allocation_id` item of the dict is not used anywhere and can be removed.

imperial_coldfront_plugin/tasks.py
```diff
@@ -220,6 +220,49 @@ def check_ldap_consistency() -> list[Discrepancy]:
...
+        return []
+
+    discrepancies: list[Discrepancy] = []
+    allocations = HX2Allocation.objects.filter(
+        resources__name="HX2",
+        status__name="Active",
+        project__projectattribute__proj_attr_type__name="Group ID",
+    ).distinct()
+
+    ldap_groups = ldap_group_member_search(f"{settings.LDAP_HX2_SHORTNAME_PREFIX}*")
```
cc-a, line 239: Looks like there is enough common logic from this point onwards between this and `check_ldap_consistency` to pull it out into a helper function.

imperial_coldfront_plugin/tasks.py
```diff
@@ -178,7 +178,7 @@ def create_rdf_allocation(form_data: AllocationFormData) -> int:
 
 
 def check_ldap_consistency() -> list[Discrepancy]:
```
cc-a:
```suggestion
def check_rdf_ldap_consistency() -> list[Discrepancy]:
```

tests/test_emails.py
```diff
@@ -85,3 +86,42 @@ def test_send_fileset_not_found_notification():
...
+
+    # Default with no source uses RDF:
+    send_discrepancy_notification(discrepancies)
+
+    assert len(mail.outbox) == 1
+    assert "RDF" in mail.outbox[0].subject
+    assert "RDF" in mail.outbox[0].body
+
+
+@override_settings(ADMINS=[("Name", "admin@email.com")])
```
cc-a, line 111: I'd somehow missed that this exists. May have to review where else we can use it.

cc-a (changes requested):
Refactor of the task tests looks good. One comment is that as they've been grouped into the classes the names of the individual test methods can be shortened.

imperial_coldfront_plugin/tasks.py
```diff
@@ -206,16 +199,51 @@ def check_ldap_consistency() -> list[Discrepancy]:
...
+
+def check_rdf_ldap_consistency() -> list[Discrepancy]:
+    """Check the consistency of LDAP groups with the RDF Active allocations."""
+    if not settings.LDAP_ENABLED:
+        return []
+
+    allocations = RDFAllocation.objects.filter(
+        resources__name="RDF Active",
+        status__name="Active",
+        allocationattribute__allocation_attribute_type__name="Shortname",
```
cc-a: Not strictly related to this PR but I think this particular line can probably go (from here and the HX2Allocation below).

cc-a (approved):
Thanks @sallymatson. Looks good. One tiny tweak, not actually strictly related to this PR.

imperial_coldfront_plugin/tasks.py
sallymatson: Okay to clarify, you mean just line 219 here (and same below) right? Last line of query

cc-a: Yep!

PR #372, opened 2026-04-10
336 chargin rates

### LINKED ISSUE #336: Store charging rates for different service offerings
We need a new setting that allows us to store the charging rates for credits consumed by different services.

This should take the form of a dictionary mapping a service name e.g. "rdf_active" to a charging rate. As charging rates will have different units we should look at representing this properly using an appropriate library like [pint](https://pint.readthedocs.io/en/stable/getting/tutorial.html). This will enable us to, for instance, store a charging rate for rdfactive as credits per TB per year.

Sahil590:
# Description
Added a setting for charging rates. 

Fixes #336 
## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/settings.py
+++ b/imperial_coldfront_plugin/settings.py
@@ -198,3 +198,4 @@ SERVICE_CHARGING_RATES: dict[str, pint.Quantity[int]] = {
     "rdf_active": ENV.int("RDF_ACTIVE_CHARGING_RATE", default=50)
     / (ureg.terabyte * ureg.year),
 }
+"""Charging rates for services as Pint Quantities in TB/year."""
--- a/imperial_coldfront_plugin/settings.py
+++ b/imperial_coldfront_plugin/settings.py
@@ -6,7 +6,10 @@ These are imported into the project level settings by the Coldfront plugin mecha
 from pathlib import Path
 from string import ascii_lowercase, digits
 
+import pint
 from coldfront.config.env import ENV
+
+from imperial_coldfront_plugin.units import ureg
 
 from .acl import ACL, ACLEntry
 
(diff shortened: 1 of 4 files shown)


imperial_coldfront_plugin/settings.py
```diff
@@ -189,3 +192,10 @@ def string_to_gid_ranges(gid_ranges_str: str) -> list[range]:
 
 ENABLE_USER_GROUP_CREATION = ENV.bool("ENABLE_USER_GROUP_CREATION", default=False)
 """Feature flag to enable or disable creation of user groups for allocations."""
+
+
+SERVICE_CHARGING_RATES: dict[str, pint.Quantity[int]] = {
+    "rdf_active": ENV.int("RDF_ACTIVE_CHARGING_RATE", default=50)
+    * ureg.credit
```
cc-a: On reflection I'm not sure we actually need a dedicated credit unit. Units can just be `(TB.year)^-1`

imperial_coldfront_plugin/settings.py
```diff
@@ -189,3 +192,10 @@ def string_to_gid_ranges(gid_ranges_str: str) -> list[range]:
...
+    / (ureg.terabyte * ureg.year),
+}
```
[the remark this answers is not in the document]
cc-a: Yeah, a docstring would be good.

imperial_coldfront_plugin/units.py
```diff
@@ -0,0 +1,6 @@
+"""Pint unit registry with custom units for the Imperial Coldfront plugin."""
+
+import pint
+
+ureg: pint.UnitRegistry[float] = pint.UnitRegistry()
+ureg.define("credit = [credit]")
```
cc-a: This can go I think.

cc-a (approved):

imperial_coldfront_plugin/units.py
Sahil590: I have removed it and changed the setting to be 'pint.unit'

cc-a (commented):
Thanks, seems to work fine like that.

PR #373, opened 2026-04-10
Add workflow for admin HX2 allocation creation.

### LINKED ISSUE #337: Create admin form for granting cluster access
This should look similar to the current `add_rdf_storage_allocation` view. Presenting a form to an admin where they select the group to grant access to, which cluster (hx2 or hx3).

Processing the form should create a new `HX2Allocation` using the `create_hx2allocation` method from #352.

sallymatson:
# Description

Creates an admin page to give a group access to HX2. This involves...
- Creating a view for both the form and the success/fail page
- Adding the URL routing 
- Adding a task that calls `HX2Allocation.objects.create_hx2allocation` - i.e. the `HX2Allocation` manager 
- Adding the templates

This workflow is only slightly different from `RDFAllocation` because there are fewer fields (only hx2/hx3 choice and the group are expose to the user). 

There is some logic to differentiate between HX2 and HX3. How it works in practice is that the admin user can select which resource they want to give the group access to on the template; the HX3 option is currently disabled. The view verifies that only the HX2 option is allowed (otherwise returns an error) and then goes through the process for creating a HX2 allocation. This is also included in the URL, which makes it more clear on the success/failure page what resource is being talked about. Ultimately it'll be quite easy to add HX3 to the workflow - just remove the diabled tag in the form and add an HX3 "if" statement in the `create_hx_allocation` task which routes to its specific manager. 

Fixes #337 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -400,6 +400,13 @@ class HXAllocationForm(forms.Form):
         return resource_type
 
 
+class HXAllocationFormData(TypedDict):
+    """Structure for holding cleaned HX allocation form data with types."""
+
+    resource_type: str
+    project: ICLProject
+
+
 class HX2TermsAndConditionsForm(forms.Form):
     """Form for accepting terms and conditions."""
 
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -5,7 +5,7 @@ This module contains form classes used for research group management.
 
 from collections.abc import Iterable
 from datetime import date, timedelta
-from typing import TYPE_CHECKING, Any, TypedDict
+from typing import TYPE_CHECKING, Any, ClassVar, TypedDict
 
 from coldfront.core.allocation.models import AllocationAttribute
 from coldfront.core.project.forms import ProjectAddUsersToAllocationForm
@@ -131,6 +131,13 @@ class AllocationFormData(TypedDict):
     size: int
     dart_id: str
     allocation_shortname: str
+
+
+class HXAllocationFormData(TypedDict):
+    """Structure for holding cleaned HX allocation form data with types."""
+
+    project: ICLProject
+    resource_type: str
 
 
 class RDFAllocationForm(forms.Form):
@@ -374,3 +381,27 @@ class CreditTransactionForm(forms.ModelForm["CreditTransaction"]):
         widget=forms.Textarea(attrs={"rows": 3}),
         help_text="Description of the transaction",
     )
+
+
+class HXAllocationForm(forms.Form):
+    """Form for creating a new HX2 allocation."""
+
+    project: forms.ModelChoiceField[ICLProject] = forms.ModelChoiceField(
+        queryset=ICLProject.objects.filter(status__name="Active"),
+        widget=_js_select_widget(),
+    )
+
+    HX_CHOICES: ClassVar[list[tuple[str, str]]] = [("hx2", "HX2"), ("hx3", "HX3")]
+    resource_type = forms.ChoiceField(
+        choices=HX_CHOICES,
+        # Remove the widget when HX3 is active, along with the javascript in the
+        # template and the clean function below.
+        widget=forms.Select(attrs={"data-disabled-options": "hx3"}),
+    )
+
+    def clean_resource_type(self) -> str:
+        """Validate that the selected HX type is available."""
+        resource_type = self.cleaned_data["resource_type"]
+        if resource_type == "hx3":
+            raise ValidationError("HX3 allocations are not currently available.")
+        return resource_type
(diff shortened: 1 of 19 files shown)


imperial_coldfront_plugin/templates/imperial_coldfront_plugin/hx_allocation_form.html
```diff
@@ -0,0 +1,37 @@
+{% extends "common/base.html" %}
+{% load crispy_forms_tags %}
+{% block content %}
+  <div class="container mt-4">
+    <div class="card">
+      <div class="card-header">
+        <h2>HX Access Request</h2>
```
cc-a:
```suggestion
        <h2>Grant Cluster Access</h2>
```

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/hx_allocation_form.html
```diff
@@ -0,0 +1,37 @@
...
+      </div>
+      <div class="card-body">
+        <form method="post"
+              id="hx_allocation_form"
+              departments-url="{% url 'imperial_coldfront_plugin:load_departments' %}">
```
cc-a: What's this for?

cc-a (changes requested):
Thanks for this @sallymatson. Looking good. The main thing I think is that we probably need the background task off loading for this. We use it on the rdf allocation side of things because of the very slow GPFS API interactions. Whilst formally it might be a good idea in this case, in practice the LDAP interactions are very fast and we've generally been accepting that these happen inside the request response cycle.

sallymatson:
> Thanks for this @sallymatson. Looking good. The main thing I think is that we probably don't need the background task off loading for this. We use it on the rdf allocation side of things because of the very slow GPFS API interactions. Whilst formally it might be a good idea in this case, in practice the LDAP interactions are very fast and we've generally been accepting that these happen inside the request response cycle.

Okay to clarify we do or don't need the offloading? I assume don't from the rest of this comment?

cc-a:
> Okay to clarify we do or don't need the offloading? I assume don't from the rest of this comment?

:facepalm: Yeah, we don't.

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/hx_allocation_form.html
sallymatson: Leftover from when I copied the rdf html - will delete

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/hx_allocation_task_result.html
```diff
@@ -0,0 +1,28 @@
...
+        <h2 class="text-justify">{{ resource_type|upper }} Access Successfully Granted for {{ group_id }}</h2>
+      {% else %}
+        <h2 class="text-justify">{{ resource_type|upper }} Access Request Failed</h2>
+        <p>
+          In the event of failure the creation process will attempt to clean up any partially created or configured resources to keep Coldfront and Active Directory. However these systems should be manually checked to ensure this is the case.
+        </p>
+        <p>
+          The following steps are recommended:
+          <ul>
+            <li>Ensure that the AD group '{{ resource_type|upper }}-{{ group_id }}' does not exist.</li>
```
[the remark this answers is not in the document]
sallymatson: The name should be correct on production and we aren't passing in the allocation variable

imperial_coldfront_plugin/forms.py
```diff
@@ -376,6 +376,37 @@ class Meta:
...
 
 
+class HXAllocationForm(forms.Form):
+    """Form for creating a new HX2 allocation."""
+
+    project: forms.ModelChoiceField[ICLProject] = forms.ModelChoiceField(
+        queryset=ICLProject.objects.filter(status__name="Active"),
+        widget=_js_select_widget(),
+    )
+
```
cc-a, line 386: You can avoid the casting in views.py with the below.

```suggestion
    cleaned_data: HXAllocationFormData
```

imperial_coldfront_plugin/tasks.py
```diff
@@ -178,6 +178,43 @@ def create_rdf_allocation(form_data: AllocationFormData) -> int:
...
+        )
+
+        logger.info(f"Created HX2 allocation with pk {hx_allocation.pk}")
+
+    else:
+        raise ValueError(f"Invalid HX resource type: {resource_type}")
+
+    return hx_allocation.pk
+
+
```
cc-a: I think we can probably drop this task and just have the logic in the view function.

imperial_coldfront_plugin/templates/imperial_coldfront_plugin/hx_allocation_task_result.html
```diff
@@ -0,0 +1,6 @@
+{% extends "common/base.html" %}
+{% block content %}
+  <div class="mb-3">
+    <h2 class="text-justify">{{ resource_type|upper }} Access Successfully Granted for {{ group_id }}</h2>
```
cc-a, line 4: Let's add a link to the detail view for the group/project.

cc-a (changes requested):
Looking good. Just a couple more suggested tweaks.

cc-a (approved):

PR #382, opened 2026-04-16
add credit transaction to admin

Sahil590:
# Description
Added `CreditTransaction` model to admin

Fixes #351 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/admin.py
+++ b/imperial_coldfront_plugin/admin.py
@@ -6,7 +6,7 @@ from .models import CreditTransaction
 
 
 @admin.register(CreditTransaction)
-class CreditTransactionAdmin(admin.ModelAdmin):
+class CreditTransactionAdmin(admin.ModelAdmin[CreditTransaction]):
     """Admin configuration for CreditTransaction model."""
 
     list_display = ("authoriser", "project", "timestamp", "amount", "description")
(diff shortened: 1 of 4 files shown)


imperial_coldfront_plugin/admin.py
```diff
@@ -1 +1,12 @@
 """Django admin configuration."""
+
+from django.contrib import admin
+
+from .models import CreditTransaction
+
+
+@admin.register(CreditTransaction)
+class CreditTransactionAdmin(admin.ModelAdmin):
```
cc-a: Mypy fix.

```suggestion
class CreditTransactionAdmin(admin.ModelAdmin[CreditTransaction]):
```

cc-a (changes requested):
Ugh. So this is more of a pain than I thought. I've suggested the change that will keep mypy happy but that actually breaks running things normally because the monkey patching needed for django-mypy doesn't happen early enough. The django-mypy docs suggests that the monkey patching should be done in settings rather than `apps.py` and moving it over does seem to work. You also need to add the monkeypatch command to `makemigrations.py` however as that doesn't use the plugin settings file.

Sahil590:
I added the command to the `makemigrations.py` but still getting the mypy error

cc-a:
See the bit of my comment about moving the monkeypatch from apps.py to settings.py.

cc-a (approved):

PR #384, opened 2026-04-17
Update CI workflow that was allowing unreproducible dependencies

cc-a:
# Description

#372 added pint as a project dependency. However pint was not also added to the dev and doc requirements files. The standard CI build and test workflow installs dependencies from `dev-requirements.txt` but the workflow was not failing despite some necessary dependencies being missing from this file. This was because of the following line in the CI:

`run: pip install -r dev-requirements.txt && pip install -e .`

This installs the development requirements and then attempts an editable install of the package. As was, if there are dependencies missing from `dev-requirements.txt` that are present in the pyproject.toml these will be installed as part of the second step. However those packages will not be installed to correct pinned versions. This allows for the CI build environment to be unreproducible. Adding the `--no-deps` flag prevents this behaviour of pip and means CI runs will fail if dependencies are missing from `dev-requirements.txt`.

Also adds `pint` and it's dependencies to to the dev and doc requirements files.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
While I definitely cannot claim expertise in ci/dependencies, all makes sense w the description

PR #387, opened 2026-04-20
Hide expiry banners for non-rdf allocations

### LINKED ISSUE #377: Hide allocation status banners for non-rdf allocations
Currently if an hx2 allocation has a non-Active status this triggers the display logic in the `imperial_coldfront_plugin/templates/imperial_coldfront_plugin/overrides/allocation_detail.html`. Those banners are intended only for rdf allocations so the rendering logic should check for the allocation type.

cc-a:
# Description

#320 added banners displayed at the top of the allocation detail page depending on the status of the allocation. Those banners only make sense when applied to RDF allocations however. At the time those were the only sorts of allocation in use. Now that we have HX2 allocations as well we need extra logic to only show them when relevant.

Fixes #377 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1243,6 +1243,19 @@ class TestAllocationDetailBanners:
             },
         )
 
+    def test_banner_not_displayed_for_non_rdf_allocations(
+        self, request_, settings, hx2_allocation
+    ):
+        """Test that no banner is displayed for non-RDF allocations."""
+        expired_status, _ = AllocationStatusChoice.objects.get_or_create(name="Expired")
+        hx2_allocation.status = expired_status
+        response = self._render_allocation_detail(request_, hx2_allocation, settings)
+        soup = BeautifulSoup(response.content, "html.parser")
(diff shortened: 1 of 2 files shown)


tests/test_views.py
```diff
@@ -1243,6 +1243,19 @@ def _render_allocation_detail(self, request_, rdf_allocation, settings):
...
         )
 
+    def test_banner_not_displayed_for_non_rdf_allocations(
+        self, request_, settings, hx2_allocation
+    ):
+        """Test that no banner is displayed for non-RDF allocations."""
+        expired_status, _ = AllocationStatusChoice.objects.get_or_create(name="Expired")
+        hx2_allocation.status = expired_status
+        response = self._render_allocation_detail(request_, hx2_allocation, settings)
+        soup = BeautifulSoup(response.content, "html.parser")
```
SaranjeetKaur, line 1253: I still need to read about BeautifulSoup and see how it works :)

SaranjeetKaur (approved):

PR #389, opened 2026-04-22
Add near expiry banner

### LINKED ISSUE #334: Add banner to allocation detail page if near to end_date
Currently we have banners that display once an allocation is expired but we should also have one when an allocation is still active but approaching expiry. This should be displayed once the first expiry warning notification email has been sent and should show the number of days until expiry.

SaranjeetKaur:
# Description

Closes #334 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


tests/test_views.py
```diff
@@ -1459,6 +1459,21 @@ def test_no_banner_for_active_allocation(self, request_, rdf_allocation, setting
         assert not soup.find("div", id="removed-allocation")
         assert not soup.find("div", id="archived-allocation")
 
+    def test_banner_for_near_expiry_allocation(
```
cc-a, line 1462: Let's add a dedicated test case for when a banner should not be displayed in this class.

cc-a (approved):

PR #403, opened 2026-05-12
341 credit allocation on creation

Sahil590:
# Description
Added a checkbox to create a credit transaction when requesting RDF allocation. This will calculate the cost of the allocation in credits and then check to see if the project has enough credits if there is it will remove the credits from the project. If there are not enough credits it will state how many are needed. 

Fixes #341 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -190,11 +190,11 @@ class RDFAllocationForm(forms.Form):
     )
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
-        """Initialise form and remove auto-credit fields when feature is disabled."""
+        """Initialise form and hide auto-credit fields when feature is disabled."""
         super().__init__(*args, **kwargs)
         if not settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT:
-            self.fields.pop("create_credit_transaction")
-            self.fields.pop("credit_transaction_description")
+            self.fields["create_credit_transaction"].widget = forms.HiddenInput()
+            self.fields["credit_transaction_description"].widget = forms.HiddenInput()
 
     def clean_dart_id(self) -> str:
         """Validate provided Dart ID."""
@@ -233,16 +233,12 @@ class RDFAllocationForm(forms.Form):
         if not settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT:
             return cleaned_data
 
-        create_credit_transaction = bool(
-            cleaned_data.get("create_credit_transaction", False)
-        )
-        if not create_credit_transaction:
-            cleaned_data["credit_transaction_description"] = ""
+        if not cleaned_data.get("create_credit_transaction"):
             return cleaned_data
 
-        description = (cleaned_data.get("credit_transaction_description") or "").strip()
-
-        cleaned_data["credit_transaction_description"] = description
+        cleaned_data["credit_transaction_description"] = (
+            cleaned_data.get("credit_transaction_description") or ""
+        ).strip()
 
         project = cleaned_data.get("project")
         size = cleaned_data.get("size")
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -5,7 +5,7 @@ This module contains form classes used for research group management.
 
 from collections.abc import Iterable
 from datetime import date, timedelta
-from typing import TYPE_CHECKING, Any, ClassVar, TypedDict
+from typing import TYPE_CHECKING, Any, ClassVar, NotRequired, TypedDict
 
 from coldfront.core.allocation.models import AllocationAttribute
 from coldfront.core.project.forms import ProjectAddUsersToAllocationForm
@@ -22,6 +22,7 @@ from django.utils import timezone
 from imperial_coldfront_plugin.models import ICLProject
 from imperial_coldfront_plugin.utils import (
     get_allocation_shortname,
+    get_rdf_allocation_credit_projection,
 )
 
 from .dart import DartIDValidationError, validate_dart_id
@@ -132,6 +133,8 @@ class AllocationFormData(TypedDict):
     size: int
     dart_id: str
     allocation_shortname: str
+    create_credit_transaction: NotRequired[bool]
+    credit_transaction_description: NotRequired[str]
 
 
 class RDFAllocationForm(forms.Form):
@@ -172,6 +175,26 @@ class RDFAllocationForm(forms.Form):
         max_length=settings.ALLOCATION_SHORTNAME_MAX_LENGTH,
         validators=[filesystem_path_component_validator],
     )
+    create_credit_transaction = forms.BooleanField(
+        required=False,
+        initial=True,
+        help_text=(
+            "Create a debit transaction from the requested storage size and duration."
+        ),
+    )
+    credit_transaction_description = forms.CharField(
+        required=False,
+        max_length=255,
+        widget=forms.Textarea(attrs={"rows": 3}),
+        help_text="Description to store on the generated credit transaction.",
+    )
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialise form and remove auto-credit fields when feature is disabled."""
+        super().__init__(*args, **kwargs)
+        if not settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT:
+            self.fields.pop("create_credit_transaction")
+            self.fields.pop("credit_transaction_description")
 
     def clean_dart_id(self) -> str:
         """Validate provided Dart ID."""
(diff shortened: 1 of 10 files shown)


imperial_coldfront_plugin/utils.py
```diff
@@ -46,6 +49,63 @@ def calculate_credit_balance(project: ICLProject) -> int:
...
+    """
+    if end_date < start_date:
+        raise ValueError("End date must be on or after start date.")
+
+    duration_days = (end_date - start_date).days + 1
+    charging_rate_per_tb_year = settings.SERVICE_CHARGING_RATES["rdf_active"].to(
+        "1 / terabyte / year"
+    )
+    charge = charging_rate_per_tb_year.magnitude * size_tb * (duration_days / 365)
+    return -round(charge)
```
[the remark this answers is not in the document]
Sahil590: used Ceil for this

imperial_coldfront_plugin/forms.py
```diff
@@ -133,6 +134,8 @@ class AllocationFormData(TypedDict):
     size: int
     dart_id: str
     allocation_shortname: str
+    create_credit_transaction: NotRequired[bool]
+    credit_transaction_description: NotRequired[str]
```
cc-a, line 137: Why `NotRequired`?

imperial_coldfront_plugin/forms.py
```diff
@@ -173,6 +176,26 @@ class RDFAllocationForm(forms.Form):
...
+        widget=forms.Textarea(attrs={"rows": 3}),
+        help_text="Description to store on the generated credit transaction.",
+    )
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialise form and remove auto-credit fields when feature is disabled."""
+        super().__init__(*args, **kwargs)
+        if not settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT:
+            self.fields.pop("create_credit_transaction")
+            self.fields.pop("credit_transaction_description")
```
cc-a: I think actually messing with the from structure is a bit messy. I'd rather just hide the fields by changing the widgets here to `HiddenInput`.

imperial_coldfront_plugin/forms.py
```diff
@@ -195,6 +218,56 @@ def clean_allocation_shortname(self) -> str:
...
+        create_credit_transaction = bool(
+            cleaned_data.get("create_credit_transaction", False)
+        )
+        if not create_credit_transaction:
+            cleaned_data["credit_transaction_description"] = ""
+            return cleaned_data
+
+        description = (cleaned_data.get("credit_transaction_description") or "").strip()
+
+        cleaned_data["credit_transaction_description"] = description
```
cc-a: It's not suitable to substitute the description field for the credit_transaction_description.

imperial_coldfront_plugin/forms.py
```diff
@@ -195,6 +218,56 @@ def clean_allocation_shortname(self) -> str:
...
+            return cleaned_data
+
+        if not settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT:
+            return cleaned_data
+
+        create_credit_transaction = bool(
+            cleaned_data.get("create_credit_transaction", False)
+        )
+        if not create_credit_transaction:
+            cleaned_data["credit_transaction_description"] = ""
```
cc-a: Why is this needed?

imperial_coldfront_plugin/utils.py
```diff
@@ -46,6 +50,63 @@ def calculate_credit_balance(project: ICLProject) -> int:
...
+        ValueError: If end_date precedes start_date.
+    """
+    if end_date < start_date:
+        raise ValueError("End date must be on or after start date.")
+
+    duration_days = (end_date - start_date).days + 1
+    charging_rate_per_tb_year = settings.SERVICE_CHARGING_RATES["rdf_active"].to(
+        "1 / terabyte / year"
+    )
+    charge = charging_rate_per_tb_year.magnitude * size_tb * (duration_days / 365)
```
cc-a: This is actually side stepping the functionality that we want to pint to provide us. We should apply units to `size_tb` and `duration_days` then take the magnitude after doing the multiplication.

imperial_coldfront_plugin/utils.py
```diff
@@ -46,6 +50,63 @@ def calculate_credit_balance(project: ICLProject) -> int:
...
+    Raises:
+        ValueError: If end_date precedes start_date.
+    """
+    if end_date < start_date:
+        raise ValueError("End date must be on or after start date.")
+
+    duration_days = (end_date - start_date).days + 1
+    charging_rate_per_tb_year = settings.SERVICE_CHARGING_RATES["rdf_active"].to(
+        "1 / terabyte / year"
+    )
```
cc-a: Not sure we need the `to` here if the size and duration have units as well.

imperial_coldfront_plugin/tasks.py
```diff
@@ -77,6 +121,16 @@ def create_rdf_allocation(form_data: AllocationFormData) -> int:
 
     logger.info("Creating initial database entries for RDF allocation.")
     with transaction.atomic():
+        if should_create_credit_transaction:
+            project = _create_rdf_allocation_debit_transaction(
```
cc-a, line 125: Why return project here?

imperial_coldfront_plugin/tasks.py
```diff
@@ -35,9 +40,43 @@
...
+    *,
+    project: ICLProject,
+    size_tb: int,
+    start_date: date,
+    end_date: date,
+    description: str,
+    authoriser: str,
+) -> ICLProject:
+    """Create a debit transaction for an RDF allocation under a project row lock."""
+    locked_project = ICLProject.objects.select_for_update().get(pk=project.pk)
```
cc-a, line 56: Why are we locking the project here when we're not making any changes to it?

imperial_coldfront_plugin/tasks.py
```diff
@@ -35,9 +40,43 @@
...
+        size_tb=size_tb,
+        start_date=start_date,
+        end_date=end_date,
+    )
+    if projected_balance < 0:
+        raise ValueError(
+            "Insufficient project credits for this allocation. "
+            f"Current balance: {current_balance} credits. "
+            f"Required debit: {-debit} credits."
+        )
```
cc-a, line 68: I think let's take this check out. There may situations where the admin team want to allow a transaction even if there aren't enough credits to cover it so I wouldn't completely forbid it.

cc-a (commented):
Still need to look at the tests but wanted to provide this feedback now.

imperial_coldfront_plugin/utils.py
```diff
@@ -31,7 +35,7 @@ def get_allocation_shortname(allocation: Allocation) -> str:
         return ""
 
 
-def calculate_credit_balance(project: ICLProject) -> int:
+def calculate_credit_balance(project: ICLProject) -> int | None:
```
cc-a: As we're using an absence of credit transactions to denote a balance of zero I think it would be better to add the logic to this function so that it returns 0 instead of None.

tests/test_forms.py
```diff
@@ -144,6 +144,105 @@ def test_rdf_allocation_end_date_initial_value(rdf_form_data, settings):
...
+
+
+def test_rdf_allocation_form_auto_credit_fields_present_when_feature_enabled(settings):
+    """Test auto-credit fields are present when the feature flag is enabled."""
+    settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT = True
+
+    form = RDFAllocationForm()
+
+    assert "create_credit_transaction" in form.fields
+    assert form.fields["create_credit_transaction"].initial is True
```
cc-a:
```suggestion
    assert form.fields["create_credit_transaction"].initial
```

tests/test_forms.py
```diff
@@ -144,6 +144,105 @@ def test_rdf_allocation_end_date_initial_value(rdf_form_data, settings):
...
+def test_rdf_allocation_rejects_when_project_has_insufficient_credit(
+    settings, rdf_form_data, project
+):
+    """Test form blocks submission when project credits are insufficient."""
+    settings.ENABLE_RDF_ALLOCATION_AUTO_CREDIT = True
+    CreditTransaction.objects.create(
+        project=project,
+        amount=0,
+        description="zero seed credit",
+    )
```
cc-a: Shouldn't be needed.

tests/test_utils.py
```diff
@@ -12,3 +18,53 @@ def test_calculate_credit_balance_returns_sum(project):
     CreditTransaction.objects.create(project=project, amount=25, description="bonus")
 
     assert calculate_credit_balance(project) == 85
+
+
+@pytest.mark.django_db
```
cc-a: Why do we need this? If it is necessary I prefer use of the `db` fixture.

tests/test_utils.py
```diff
@@ -12,3 +18,53 @@ def test_calculate_credit_balance_returns_sum(project):
...
+def test_calculate_rdf_allocation_credit_debit_uses_inclusive_days():
+    """Test debit calculation uses inclusive day count and rounding."""
+    debit = calculate_rdf_allocation_credit_debit(
+        size_tb=1, start_date=date(2026, 1, 1), end_date=date(2026, 2, 1)
+    )
+
+    assert debit == -5
```
cc-a, line 29: These tests and some others seem to be depend on the default value for `RDF_ACTIVE_CHARGING_RATE` being used. If the environment variable is set to something else this will probably cause the tests to fail. The best thing to do is to override the setting in `conftest.py` so it's guaranteed to be a fixed value.

cc-a (changes requested):
Looking pretty good. Just a few additional small points.

imperial_coldfront_plugin/forms.py
Sahil590: When the creation transaction is ticked it was preventing it from being submitted

imperial_coldfront_plugin/tasks.py
Sahil590: I added for the concurrency issue.

cc-a: What concurrency issue? The project is not being updated.

imperial_coldfront_plugin/forms.py
cc-a: `NotRequired` is only meaningful in relation to `TypedDict`s. Removing these makes no difference to the tests or mypy.

imperial_coldfront_plugin/tasks.py
Sahil590: It was pointed out here https://github.com/ImperialCollegeLondon/imperial_coldfront_plugin/pull/403#discussion_r3227620892, so I based it of this comment. Looking at it now would be unlikely a transaction occurs during the form submission.

imperial_coldfront_plugin/forms.py
cc-a: I think these can be removed as the fields are now always present and form validation always puts some values for these.

cc-a (approved):

PR #405, opened 2026-05-20
Fix integration issue with allocation form and project-add-users view

cc-a:
# Description

This PR fixes a subtle issue that prevented users from being added to projects that have only a HX2 allocation. The logic for form hiding was not fully working due to a mismatch in the logic between `ProjectAddUsersToAllocationShortnameForm.__init__` and `ProjectAddUsersSearchResultsShortnameView.post`. This would lead to invalid data being passed to the `project-add-user` view and users being failing to be added to a project.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -349,14 +349,11 @@ class ProjectAddUsersToAllocationShortnameForm(ProjectAddUsersToAllocationForm):
             allocation_choices, key=lambda x: x[1][0].lower()
         )
         allocation_choices.insert(0, ("__select_all__", "Select All"))
-        if allocation_query_set:
-            self.fields["allocation"].choices = allocation_choices_sorted
-            self.fields["allocation"].help_text = (
-                "<br/>Select allocations to add selected users to. HX2 allocations are "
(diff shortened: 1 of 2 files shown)


imperial_coldfront_plugin/forms.py
```diff
@@ -349,14 +349,11 @@ def __init__(
             allocation_choices, key=lambda x: x[1][0].lower()
         )
         allocation_choices.insert(0, ("__select_all__", "Select All"))
```
SaranjeetKaur, line 351:
```suggestion
        allocation_choices_sorted.insert(0, ("__select_all__", "Select All"))
```
Do we want to insert the "Select All" to the sorted choices instead?

cc-a: Ugh.. in theory yes. So that's a bug that comes from the Coldfront code that we've had to duplicate here. In theory yes we could fix it, in practice it's an orthogonal change that I don't want to include here.

SaranjeetKaur: Ok, cool!

SaranjeetKaur (approved):

PR #425, opened 2026-06-25
Exponential Back-off Retry for fileset creation

### LINKED ISSUE #417: Add retry with exponential back-off for fileset creation
In production a lag has been observed between creation of an AD group and GPFS being able to see that group which is a pre-requisite for fileset creation. Until now we've hackily addressed this via `GPFS_ALLOCATION_CREATION_SLEEP` which simply sleeps for a configurable period to allow AD and GPFS to sync. This is proving a bit flaky though so we should have a proper retry approach with exponential back-off and a configurable maximum timeout.

cc-a:
# Description

This PR removes the previous arbitrary length sleep before attempting fileset creation. Instead it implements an exponential backoff retry scheme using the `backoff` library with a configurable maximum timeout.

Fixes #417 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/settings.py
+++ b/imperial_coldfront_plugin/settings.py
@@ -189,8 +189,8 @@ DEPARTMENTS_IN_FACULTY = {
     "fons": ["physics", "chemistry"],
 }
 
-GPFS_ALLOCATION_CREATION_SLEEP = ENV.int("GPFS_ALLOCATION_CREATION_SLEEP", default=30)
-"""Number of seconds to sleep before creating a GPFS fileset."""
+GPFS_AD_SYNC_TIMEOUT_SECONDS = ENV.int("GPFS_AD_SYNC_TIMEOUT_SECONDS", default=600)
+"""Maximum wait time configured for exponential backoff retries for fileset creation."""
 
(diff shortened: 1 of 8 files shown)


imperial_coldfront_plugin/settings.py
```diff
@@ -189,8 +189,8 @@ def string_to_gid_ranges(gid_ranges_str: str) -> list[range]:
     "fons": ["physics", "chemistry"],
 }
 
-GPFS_ALLOCATION_CREATION_SLEEP = ENV.int("GPFS_ALLOCATION_CREATION_SLEEP", default=30)
-"""Number of seconds to sleep before creating a GPFS fileset."""
+GPFS_AD_SYNC_TIMEOUT_SECONDS = ENV.int("GPFS_AD_SYNC_TIMEOUT_SECONDS", default=600)
```
SaranjeetKaur, line 192: How did we come up with the default time 600?

SaranjeetKaur (approved):

imperial_coldfront_plugin/settings.py
cc-a: That's the sleep value that the platforms team have been having to use previously.

PR #427, opened 2026-07-02
Refine RDF credit behaviour

cc-a:
# Description

Minor refinements to the credit transaction behaviour for RDF allocations:
- create a new credit transaction type - "Credit"
- use the Storage transaction type for RDF allocations
- auto-generate the description for RDF allocation credit transactions.
- use "RCS Credits" in the UI rather than just "Credits".

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


tests/test_tasks.py
```diff
@@ -305,8 +305,9 @@ def test_success_creates_credit_transaction_when_enabled(
 
         debit_transaction = CreditTransaction.objects.get(
             project=project,
-            description="Auto debit for RDF allocation",
+            description=f"Creation of RDF allocation - {shortname}",
```
Sahil590, line 308: I thought you had removed description from the model wont this fail?

Sahil590 (approved):
LGTM. I found one issue but will approve anyway since its minor

tests/test_tasks.py
cc-a: `description` is still a field on the model I've just removed it from `RDFAllocationForm` and replaced it with an auto-generated value rather than a user-supplied one.

PR #437, opened 2026-08-07
Add support for openldap

cc-a:
# Description

A few minor accompanying tweaks to better support the new openldap development infrastructure.

PR #441, opened 2026-08-14
Creating only Coldfront objects for RDFAllocation

### LINKED ISSUE #416: Support creating only Coldfront objects for an RDFAllocation
Presently we have the view `add_rdf_storage_allocation` that is used by admins to create new RDFAllocation instances. This triggers the background task `create_rdf_allocation` which has responsibility for creating the allocation in the database and associated resources like the LDAP group and GPFS fileset.

In a disaster recovery scenario Coldfront may have previously created a fileset and/or LDAP group but may have lost the local entries in the database. In this case we want a process whereby admins can provide the necessary information for creation of the `RDFAllocation` instance. When submitted the existence of the expected ldap group and fileset are confirmed, then the Coldfront data is created as required. This can then be presented as a separate option in the admin drop down menu.

SaranjeetKaur:
# Description

@cc-a  I've used copilot to write parts of this PR. Please feel free to comment extra critically on it.

Fixes #416 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/imperial_coldfront_plugin/forms.py
+++ b/imperial_coldfront_plugin/forms.py
@@ -485,3 +485,123 @@ class HX2TermsAndConditionsForm(forms.Form):
         required=True,
         label="I accept the terms and conditions of the RCS Access Policy.",
     )
+
+
+class RecoverRDFAllocationFormData(TypedDict):
+    """Structure for holding cleaned recover RDF allocation form data with types."""
+
+    project: ICLProject
+    description: str
+    start_date: date
+    end_date: date
+    size: int
+    dart_id: str
+    allocation_shortname: str
+    gid: int
+
+
+class RecoverRDFAllocationForm(forms.Form):
+    """Form for recovering Coldfront objects for an existing RDF allocation."""
+
+    project: forms.ModelChoiceField[ICLProject] = forms.ModelChoiceField(
+        queryset=ICLProject.objects.filter(status__name="Active"),
+        widget=_js_select_widget(),
+    )
+    description = forms.CharField(widget=forms.Textarea())
+    start_date = forms.DateField(
+        initial=_todays_date,
+        widget=forms.DateInput(attrs={"type": "date"}),
+    )
+    end_date = forms.DateField(
+        initial=_initial_end_date,
+        widget=forms.DateInput(attrs={"type": "date"}),
+    )
+    size = forms.IntegerField(
+        validators=[MinValueValidator(1)], help_text="In terabytes"
+    )
+    dart_id = forms.CharField(
+        help_text="The associated DART entry.",
+        disabled=False,
+        required=False,
+        widget=forms.HiddenInput(),
+    )
+    allocation_shortname = forms.CharField(
+        help_text=(
+            "Shortname of the existing allocation to recover."
+            " Lower case letters and numbers only. Must contain between "
+            f"{settings.ALLOCATION_SHORTNAME_MIN_LENGTH} and "
+            f"{settings.ALLOCATION_SHORTNAME_MAX_LENGTH} characters."
+        ),
+        min_length=settings.ALLOCATION_SHORTNAME_MIN_LENGTH,
+        max_length=settings.ALLOCATION_SHORTNAME_MAX_LENGTH,
+        validators=[filesystem_path_component_validator],
+    )
+
+    def clean_dart_id(self) -> str:
+        """Validate provided Dart ID."""
+        dart_id = self.cleaned_data["dart_id"]
+        allocation = self.cleaned_data.get("allocation")
+        if dart_id:
+            try:
+                validate_dart_id(dart_id, allocation)
+            except DartIDValidationError as e:
+                raise ValidationError(e.args[0])
+        return dart_id
+
+    def clean_allocation_shortname(self) -> str:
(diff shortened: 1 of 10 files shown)


imperial_coldfront_plugin/forms.py
```diff
@@ -485,3 +485,123 @@ class HX2TermsAndConditionsForm(forms.Form):
...
+    description: str
+    start_date: date
+    end_date: date
+    size: int
+    dart_id: str
+    allocation_shortname: str
+    gid: int
+
+
+class RecoverRDFAllocationForm(forms.Form):
```
cc-a, line 503: For both of these classes there is a lot of overlap with the pre-existing `AllocationFormData` and `RDFAllocationForm` classes. It would be good to capture the repeated fields and logic within common base classes.

imperial_coldfront_plugin/ldap.py
```diff
@@ -271,3 +271,26 @@ def ldap_gid_in_use(gid: int, conn: Connection | None = None) -> bool:
...
+    if conn is None:
+        conn = _get_ldap_connection()
+    _, _, response, _ = conn.search(
+        settings.LDAP_GROUP_OU,
+        f"(cn={group_name})",
+        attributes=["gidNumber"],
+    )
+    if not response:
+        return None
+    return int(response[0]["attributes"]["gidNumber"])
```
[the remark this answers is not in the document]
cc-a: From manual testing this is not the case with the production system. Possibly this is a ldap vs AD thing, we'll have to see how it interacts with the ldap dev infrastructure.

imperial_coldfront_plugin/forms.py
```diff
@@ -485,3 +485,123 @@ class HX2TermsAndConditionsForm(forms.Form):
...
+            self.add_error("end_date", "End date must be on or after start date.")
+            return cleaned_data
+
+        shortname = cleaned_data.get("allocation_shortname")
+        if not shortname:
+            return cleaned_data
+
+        ldap_group_name = f"{settings.LDAP_RDF_SHORTNAME_PREFIX}{shortname}"
+
+        if settings.LDAP_ENABLED:
```
cc-a, line 585: As doing anything useful with this form explicitly depends on being able to retrieve a GID I think we do probably want to got the extra step and fail form validation entirely if ldap is not enabled rather than just not populating a gid value.

imperial_coldfront_plugin/tasks.py
```diff
@@ -233,6 +233,114 @@ def create_rdf_allocation(form_data: AllocationFormData, authoriser: str = "") -
...
+    Returns:
+        The primary key of the created Allocation.
+    """
+    logger = logging.getLogger("django-q")
+
+    storage_size_tb = form_data["size"]
+    project = form_data["project"]
+    shortname = form_data["allocation_shortname"]
+    gid = form_data["gid"]
+
```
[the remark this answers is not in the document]
cc-a: Addressed by my comment in forms.py.

imperial_coldfront_plugin/tasks.py
```diff
@@ -233,6 +233,114 @@ def create_rdf_allocation(form_data: AllocationFormData, authoriser: str = "") -
     return rdf_allocation.pk
 
 
+def recover_rdf_allocation(
```
cc-a, line 237: As with the forms this is duplicating a lot of logic from `create_rdf_allocation`, it would be good to break things down a bit into logic that could be used in both places.

cc-a (changes requested):
Thanks @SaranjeetKaur. This is looking good just needs a pass to tighten it a bit.

Another check that we should do for the department and faculty values crossed my mind but is a bit more involved so I'll write it up as a separate issue.

SaranjeetKaur:
@cc-a for the manual testing, could you please share what you had updated in the .env file (related to LDAP)? I don't think I made a note of it

cc-a:
@SaranjeetKaur to use the development ldap setup make sure you have the latest version of `develop` for the development environment. It's use is documented in the "Local LDAP" section of the documentation.

We've already had to make some minor accommodations for the LDAP vs AD differences in #437. If this is another example of where we have to make code changes just to support the local LDAP setup then it's worth thinking about the trade offs there and the best way to handle them.

imperial_coldfront_plugin/ldap.py
```diff
@@ -271,3 +271,54 @@ def ldap_gid_in_use(gid: int, conn: Connection | None = None) -> bool:
...
+
+    raw_gid = response[0].get("attributes", {}).get("gidNumber")
+    gid_text = _normalise_single_ldap_value(raw_gid)
+    if gid_text is None:
+        return None
+
+    try:
+        return int(gid_text)
+    except ValueError:
+        return None
```
SaranjeetKaur: @cc-a I've used copilot to rewrite the return response of the `ldap_get_group_gid` and the helper function `_normalise_single_ldap_value` above. It would be helpful for me to see how you do the manual testing to verify.

I've added corresponding unit tests to verify the shapes in `test_ldap.py`.

SaranjeetKaur:
@cc-a 

For testing purpose, I've created a group in coldfront with an allocation

image

Next should I delete it and then test the recover RDF allocation?

